PR #17, opened 2016-09-09
Ocd 875

dlucas1:


andlar:
The only problem I can see here is in your RELEASE_NOTES file. The issue is that your OCD-875 branch is some way behind the development branch, so if this was merged just as is, the "New Feature" you added to the RELEASE_NOTES would end up getting inserted in as part of Version 4.0.0, where it should instead be in a TBD new Version. I think you need to, on your local system, switch to your development branch, pull from upstream, switch back to OCD-875, merge in your development branch (fixing any conflicts), and then push OCD-875 again.

PR #28, opened 2016-10-24
OCD-960: remove duplicate cps in production

kekey1:
the script is very specific to the duplicate products we know about in production today. if there are more or less duplicates when we run this, it will fail. open to other suggestions!


dlucas1:
Is this awaiting a code review?

kekey1:
Ah I forgot this one was hanging out there. I believe we were waiting to merge it in so that dev could be used for testing the tx/cache/etc issue.

PR #29, opened 2016-10-27
Ocd 971

kekey1:
- Script to remove duplicate cps in production (I could swear this was already merged in but it seems that it was not). This will ONLY work on production.
- SQL code to add product ownership history table.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,13 @@
 # Release Notes
 
+## Version TBD
+_Date TBD_
+
+### Changes
+* Added script to remove three specific duplicate certified products
+
+---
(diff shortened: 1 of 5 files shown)


kekey1:
Added IF EXISTS to the DROP TABLE command. I got some errors when I tried to add DROP TRIGGER and found that the DROP TABLE command is automatically dropping the trigger. (https://www.postgresql.org/docs/8.2/static/sql-droptable.html). I had previously run the script multiple times locally and did not have issues adding the triggers.

Also added newlines to the ocd-* files and dos2unixed them (which I don't normally do so I am not sure if that's right or if I've been doing it wrong all this time!)

PR #32, opened 2016-11-16
OCD-1001: Build script to migrate legacy surveillance to new tables

dlucas1:
Please review


dlucas1:
Added Andrew and Katy as assignees (I think that will allow them to get automatic email updates from this PR?)

ocd-1001.sql
```diff
@@ -0,0 +1,126 @@
...
+
+DROP TABLE IF EXISTS generalCPs;
+CREATE TEMP TABLE generalCPs AS
+SELECT C.certified_product_id
+FROM openchpl.corrective_action_plan C
+LEFT JOIN openchpl.corrective_action_plan_certification_result CR ON C.corrective_action_plan_id = CR.corrective_action_plan_id
+WHERE C.corrective_action_plan_id IS NOT NULL AND CR.corrective_action_plan_id IS NULL;
+
+SELECT print_notice('Please run the following command to output the table generalCPs to a CSV for Jennifer to review:
+COPY (SELECT * FROM generalCPs) TO ''/generalCPs.copy;''');
```
andlar: Can this be done automatically? When deploying to PROD we try to avoid requiring any action other that loading the new `vM.N.O.sql` file to reduce downtime / make it closer to automated. As well, does this leave the function in the system? I don't think we want a one-time use function sitting in the data-model any longer than needed. Check `v6.0.0.sql` for an example of a function that was created, ran, then deleted all as part of the deployment

andlar:
I think this is good now, but also think we can't merge it in until we're confident the "target" tables are locked in

PR #34, opened 2016-11-18
OCD-1014: add meaningful_use_users BIGINT column to certified_product & certified_product_details

dlucas1:
Questions:
1. Since we are using Postgresql 9.4, the below code is not yet supported - is there a reason why we're still using an older version? Should we consider upgrading to be able to support newer features like shown below? The code below works with Postgresql 9.6 and above:
ALTER TABLE openchpl.certified_product ADD COLUMN IF NOT EXISTS meaningful_use_users BIGINT;

2. Would it be worthwhile to keep the function below instead of dropping it? There is not a very easy way (in Postgresql 9.4) of checking if a column exists and adding it. See function below:
add_column(schema_name TEXT, table_name TEXT,
column_name TEXT, data_type TEXT)

### DIFF
--- a/dev/openchpl_views.sql
+++ b/dev/openchpl_views.sql
@@ -82,6 +82,7 @@ SELECT
     a.accessibility_certified,
     a.product_additional_software,
     a.last_modified_date,
+    a.meaningful_use_users,	
     b.year,
     c.certification_body_name,
     c.certification_body_code,
--- a/dev/openchpl_views.sql
+++ b/dev/openchpl_views.sql
@@ -148,7 +148,6 @@ FROM openchpl.certified_product a
     ON a.certified_product_id = s.certified_product_id
     LEFT JOIN (SELECT testing_lab_id, name as "testing_lab_name", testing_lab_code from openchpl.testing_lab) q on a.testing_lab_id = q.testing_lab_id
     ;
-	GRANT ALL ON TABLE openchpl.certified_product_details TO openchpl;
 
 -- ALTER VIEW openchpl.certified_product_details OWNER TO openchpl;
 
--- a/dev/openchpl_views.sql
+++ b/dev/openchpl_views.sql
@@ -148,6 +148,7 @@ FROM openchpl.certified_product a
     ON a.certified_product_id = s.certified_product_id
     LEFT JOIN (SELECT testing_lab_id, name as "testing_lab_name", testing_lab_code from openchpl.testing_lab) q on a.testing_lab_id = q.testing_lab_id
     ;
+	GRANT ALL ON TABLE openchpl.certified_product_details TO openchpl;
 
 -- ALTER VIEW openchpl.certified_product_details OWNER TO openchpl;
 
--- a/ocd-1014.sql
+++ b/ocd-1014.sql
@@ -1,0 +1,272 @@
+﻿CREATE OR REPLACE FUNCTION add_column(schema_name TEXT, table_name TEXT, 
+column_name TEXT, data_type TEXT)
+RETURNS BOOLEAN
+AS
+$BODY$
+DECLARE
+  _tmp text;
+BEGIN
+
+  EXECUTE format('SELECT COLUMN_NAME FROM information_schema.columns WHERE 
+    table_schema=%L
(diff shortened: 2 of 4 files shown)


andlar:
1. We're using an old version because that's what was current when we started, and we don't have the time to go back and upgrade anything right now. That's why the UI is still on an old version of Angular as well. I don't know that upgrading Postgres would be "complicated", but it definitely would need to be done with care, and right now there are too many other things to do.

2. I'd rather not keep the function there, even though it could be useful in the future. If we do need it again I'd rather just pull it back in (from git history) and use it then. Adding functionality to our data-model for things that aren't strictly part of CHPL, but instead are for supporting CHPL, doesn't seem like a good idea to me, though I'm not 100% convinced either way.

In other news, I don't think we want this PR going to the development branch. This story falls into the category of specifically timed release window stories, and so can't get into our regular pipeline for development. I'm not sure if this story can go in with the surveillance story, or if it needs a separate branch; possibly "decertifications" or something like that, but will need to look at the upcoming release schedule and make a plan

dlucas1:
Thanks for the response on the questions, and good point on the release timeline; I didn't realize this was something that should probably wait for a future release.

Any code review specific comments?

andlar:
The only other comment I had was that I think you might want to wait on this PR until the API is ready to go. The API & Data-Model work so closely together that I'd be surprised to have a change to the data-model without a corresponding change to the API. Especially if something that happened in the data-model breaks something in the API (which I don't think there is in this one, but...)

dlucas1:
Thanks for your feedback, Andrew. I'll work on the API changes before finalizing this PR; in the meantime, I would also like to receive code review feedback from Katy, as she probably has the most familiarity with the data-model & API.

kekey1:
I don't have much to add here. I agree with Andrew on not upgrading the db at this time. Your function could be useful in the future certainly. Could you also just drop the column first (IF EXISTS) and then add the column each time in the script to make it able to run repeatedly?

I suspect the search and/or getting certified details does not work when a change is made to that details_view in the db.  DB and API (and often UI) should go out together. Also, even if we PRed the DB changes there is nothing to be QAed so we really need the whole user story to be ready for testing before merging the PR.

dlucas1:
@kekey1 the problem with dropping the column if it exists is the following scenario:
Let's say it does exist, and someone is running this script in the future. If he runs the script and drops the column, he would lose all data in that column. My solution prevents that from happening. Again, it would be simple if we had Postgres 9.6, but I understand that upgrading our software is not as high a priority as delivering.

Agreed on DB + API going out together. I am working on the API changes now. Thanks for the feedback!

andlar:
We're not worried about losing data now. The purpose of the "repeatable" part of the OCD-1014.sql file is that we can run it over and over on DEV/STG as required. Once it goes to PROD it won't be run more than once

dlucas1:
@andlar understood; I will keep that in mind for the future as it could reduce development complexity/time.

dev/openchpl_views.sql
```diff
@@ -148,6 +148,7 @@ FROM openchpl.certified_product a
     ON a.certified_product_id = s.certified_product_id
     LEFT JOIN (SELECT testing_lab_id, name as "testing_lab_name", testing_lab_code from openchpl.testing_lab) q on a.testing_lab_id = q.testing_lab_id
     ;
+	GRANT ALL ON TABLE openchpl.certified_product_details TO openchpl;
```
andlar: This line isn't required. When running the scripts that uses the files in the `/dev` directory it runs the `grant-all.sql` file, which does the required permission granting

ocd-1014.sql
```diff
@@ -269,4 +269,5 @@ CREATE OR REPLACE VIEW openchpl.certified_product_details AS
 		    testing_lab.name AS testing_lab_name,
 		    testing_lab.testing_lab_code
 		   FROM openchpl.testing_lab) q ON a.testing_lab_id = q.testing_lab_id;
+		   GRANT ALL ON TABLE openchpl.certified_product_details TO openchpl;
```
andlar: This one _is_ required, as the `grant-all` script doesn't run in PROD

PR #41, opened 2016-12-05
OCD-938: add acb_is_deleted to certified_product_details view

dlucas1:


dlucas1:
There is no OCD-938 file left. It's a little deceiving since I renamed the existing OCD-938 to OCD-938-1014.sql, put the OCD-1014 changes in the OCD-938-1014.sql file, then deleted the OCD-1014.sql file.

PR #42, opened 2016-12-07
OCD-1001: updated requirement_type logic & surveillance_nonconformity mappings and logic

dlucas1:
* updated query for surveillance_requirement.type_id
a. sets type_id = 1 (Certified Product) when certification_criterion_id is not null
b. sets type_id = 2 (Transparency or Disclosure Requirement) when certification_criterion_id is null
* updated query for surveillance_nonconformity (mapping changes, join changes)

### DIFF
--- a/ocd-1001.sql
+++ b/ocd-1001.sql
@@ -4,18 +4,30 @@ TRUNCATE openchpl.surveillance CASCADE;
 -- Build temp staging tables for each destination table
 DROP TABLE IF EXISTS surveillance;
 CREATE TEMP TABLE surveillance AS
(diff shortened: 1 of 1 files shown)


ocd-1001.sql
```diff
@@ -4,18 +4,30 @@ TRUNCATE openchpl.surveillance CASCADE;
 -- Build temp staging tables for each destination table
 DROP TABLE IF EXISTS surveillance;
```
andlar: Do these tables need to be in a schema? Even if they're temp tables, will they have issues on the deployment machines without a schema?

dlucas1: Based on the fact that everything worked last time (except for the function name), no they do not. I'd rather not change something that's not broken.

PR #48, opened 2016-12-29
OCD-1039: add indexes to improve performance

dlucas1:
Note that this will not be merged/deployed to dev and staging until after we deploy the current release to production.


v-next.sql
```diff
@@ -801,4 +801,275 @@ FROM openchpl.certified_product a
...
+    CREATE INDEX ix_test_procedure ON openchpl.test_procedure (test_procedure_id, deleted);
+END IF;
+
+IF to_regclass('openchpl.ix_test_standard') IS NULL THEN
+    CREATE INDEX ix_test_standard ON openchpl.test_standard (test_standard_id, deleted);
+END IF;
+
+END$$;
+
+/* DROP INDEXES SCRIPT:
```
kekey1: Should this be uncommented and run before adding all the indexes so we can run this file multiple times?

andlar: I think there was an "IF" statement before each of the indices, but am not sure exactly how that works, without trying it myself. Definitely need to be able to run the file multiple times without running into issues

kekey1: Oh I see. Maybe all the drops don't need to be in there at all then?

dlucas1: Correct the following allows the SQL to be run multiple times: 
IF to_regclass('indexName') IS NULL THEN CREATE INDEX index_name ON schema.table_name (col1, col1, etc...); 
END IF;

However, if we remove a column that is being indexed and run this script, then the script would throw an error at that line.

dlucas1:
@andlar do I not have the necessary permissions to create a branch? I'm not seeing an option to do it on github.

andlar:
Not sure @dlucas1 . I just created a branch called "next-version". Still haven't had coffee; could probably have come up with a better name, but for now...

https://github.com/chpladmin/chpl-data-model/tree/next-version

v-next.sql
andlar: Agree with @kekey1. The commented out stuff isn't necessary. WRT the column being removed, we shouldn't have an issue with that, because part of the development process includes re-running this file during dev, so those errors would be noticed and fixed, right? ;)

dlucas1: The reason I added the commented section to drop the indexes is in case there is a bug due to an index or a significant performance issue and we need to revert the indexes quickly--we'll have the script ready to go in this case. Given that reasoning, please let me know if I should remove the commented section and I will do it.

dlucas1:
@andlar thanks for creating the next-version branch. Do you mind creating one for CHPL-API also?

v-next.sql
andlar: Good plan, but the thing I worry about is that, if we do end up removing (or adding) columns in the future, the commented out part will be obsolete. I'm okay with keeping the structure of the "remove indices" part, but maybe change the commented out part to something like:
`in case an index needs to be removed, run code like this:
REMOVE INDEX from SOMEWHERE LIKE a thing
(or whatever the actual code looks like)`

That way we'll have the format saved, but won't have to keep code up to date (especially code that isn't even being run)

PR #63, opened 2017-02-27
OCD-1249: add user_permission_id to surveillance to track surveillance authority

dlucas1:


dev/openchpl_data-model.sql
```diff
@@ -2308,14 +2308,18 @@ CREATE TABLE openchpl.surveillance (
 	creation_date timestamp NOT NULL DEFAULT NOW(),
 	last_modified_date timestamp NOT NULL DEFAULT NOW(),
 	last_modified_user bigint NOT NULL,
+	user_permission_id integer NOT NULL,
```
andlar: The user_permission_id column in the user permission table is type `bigint`, right? Is there any reason this column is of type `integer`? I know it's extremely unlikely we'd need to take advantage of the difference between integer and bigint, but it seems odd that they wouldn't match

RELEASE_NOTES.md
```diff
@@ -1,5 +1,13 @@
 # Release Notes
 
+## Version TBD
+_27 March 2017_
```
andlar: While it's likely this will be the correct date, it shouldn't be in here now

PR #71, opened 2017-03-29
Development

kekey1:


v-next.sql
```diff
@@ -24,8 +24,6 @@ CREATE TABLE openchpl.vendor_status_history (
 GRANT ALL ON TABLE openchpl.vendor_status_history TO openchpl;
 GRANT ALL ON SEQUENCE openchpl.vendor_status_history_vendor_status_history_id_seq TO openchpl;
 
-DROP TRIGGER vendor_status_history_audit on openchpl.vendor_status;
-DROP TRIGGER vendor_status_historys_timestamp on openchpl.vendor_status;
 CREATE TRIGGER vendor_status_history_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.vendor_status_history FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
```
andlar: If `v-next.sql` is run more than once, will this create more than one trigger?

kekey1: No because when the vendor_status_history table is dropped (above) the associated trigger is dropped as well. I had DROP...IF EXISTS... before and it kept telling me that it didn't exist no matter how many times I ran it!!

andlar: Gotcha. Hadn't thought about the table dropping triggering the dropped triggers, and the section of code showing here didn't show the "drop table" part which might have made me think of it

PR #73, opened 2017-04-20
OCD-1456: adjust deleted and survReq result values from CAP migration

kekey1:


dlucas1:
Should there be a release note comment for the bug fix?

kekey1:
Yes there should!! :)

dlucas1:
It may be a good idea to wrap these changes into a transaction in order to meet the four "ACID" properties.

dlucas1:
While using subqueries does get the job done, it might be cleaner to use something like as follows (note: throw away select code after testing):

BEGIN;
SELECT *
FROM openchpl.surveillance_nonconformity sn
LEFT JOIN openchpl.surveillance_requirement sr ON sn.surveillance_requirement_id = sr.id
WHERE sr.result_id = 2;
-- expect 83 before update

UPDATE openchpl.surveillance_requirement sr
SET result_id = 1
FROM openchpl.surveillance_nonconformity sn
WHERE sn.surveillance_requirement_id = sr.id AND sr.result_id = 2;

SELECT *
FROM openchpl.surveillance_nonconformity sn
LEFT JOIN openchpl.surveillance_requirement sr ON sn.surveillance_requirement_id = sr.id
WHERE sr.result_id = 2;
-- expect 0 after update
ROLLBACK;

PR #81, opened 2017-05-17
Test Standards fixes

kekey1:
Please look this over carefully even though I've run it multiple times myself! The test standards were worse off than the test functionality. Some of the preloaded data was wrong to begin with for 2014 so I'm updating that first. Then, I'm trying to figure out which edition (or both) user-entered test standards belong to. Finally, I am looking for any mismatches (ex: a user picked a 2014 test standard for a 2015 listing) and fixing those. I THINK that gets everything.


v-next.sql
```diff
@@ -0,0 +1,493 @@
...
+UPDATE openchpl.certification_result_test_standard 
+SET test_standard_id = 
+	(SELECT MIN(test_standard_id) FROM openchpl.test_standard WHERE number = '170.205(a)(1)' and test_standard_id <= 47)
+WHERE test_standard_id IN (SELECT test_standard_id FROM openchpl.test_standard WHERE number = '170.205(a)(1)' and test_standard_id <= 47);
+--delete any other test standards with this number
+UPDATE openchpl.test_standard 
+SET deleted = true, last_modified_user = -1
+WHERE number = '170.205(a)(1)' 
+AND test_standard_id > (SELECT MIN(test_standard_id) FROM openchpl.test_standard WHERE number = '170.205(a)(1)' and test_standard_id <= 47) 
+AND test_standard_id <= 47;
```
andlar: ~~Why the `<= 47` part? If that's the first ID for 170.205(a)(1), then why the same number for the following changes? (170.205(a)(2) et al)~~
Nevermind; that's the breakpoint before 2015 standards

v-next.sql
```diff
@@ -0,0 +1,493 @@
...
+					AND cp.certification_edition_id = 3)
+					LOOP
+						UPDATE openchpl.certification_result_test_standard
+						SET test_standard_id = 
+							(SELECT test_standard_id 
+							FROM openchpl.test_standard 
+							WHERE number = userEnteredTestStandard.number AND certification_edition_id = 3)
+						WHERE certification_result_test_standard_id = mismatchTestStandardId;
+					END LOOP;
+				ELSE
```
andlar: How can this happen? If it was Case 0, then it was handled above. Same with 1 and 2. Is this only possible if it was actually Case 3? If it's actually something other than 0, 1, or 2, I feel like the error message should be different

PR #83, opened 2017-06-07
OCD-1528: bulk change ICSA listings disclosure URLs

dlucas1:


v-next.sql
```diff
@@ -95,4 +95,14 @@ atl.testing_lab_name, acb.certification_body_name,prac.practice_type_name,versio
...
+-- should update 1,341 listings' Tranparency Attestation URL
```
andlar: Can you have this output the number of listings that actually change as well as the number of expected ones?

PR #95, opened 2017-07-26
OCD-1694: Add "subscriptions.sql" file; update procedure document

andlar:


maint/subscriptions.sql
```diff
@@ -0,0 +1,36 @@
...
+insert into openchpl.notification_recipient (email, last_modified_user) values
+    ('amore@ainq.com', -2)
+;
+
+/*
+ * create non-acb subscriptionss
+ */
+insert into openchpl.notification_type_recipient_map (recipient_id, notification_type_id, last_modified_user) values
+    ((select id from openchpl.notification_recipient where email = 'amore@ainq.com' and deleted = false), 3, -2)
+    ,((select id from openchpl.notification_recipient where email = 'amore@ainq.com' and deleted = false), 4, -2)
```
kekey1: If it would make the sql too verbose, I don't think it's a big deal, but it would be nice to see a nested select id from openchpl.notification_type where name = 'ONC Something Report' instead of just the id 3, 4, etc so to someone reading this it's very clear what subscriptions are getting added.

PR #108, opened 2017-09-26
Ocd 1730

brianlindsy:


dev/openchpl_data-model.sql
```diff
@@ -27,6 +27,24 @@ CREATE TYPE openchpl.attestation as enum('Affirmative', 'Negative', 'N/A');
 CREATE TYPE openchpl.validation_message_type as enum('Error', 'Warning');
 CREATE TYPE openchpl.job_status_type as enum('In Progress', 'Complete', 'Error');
 
+CREATE TABLE openchpl.chart_data(
+	chart_data_id bigserial NOT NULL,
```
kekey1: A lot of the original tables use these long id column names but in all the newer ones I've been creating for a while I just call it "id". I'm not sure if we'll ever get to go back and change the old ones but could you make these match the newer style? The reason is that PSQL automatically creates a sequence for bigserial types using the pattern table_column_seq, so all the sequences with this longer style id have names like "chart_data_chart_data_id_seq" which is awkward. If we just call the column "id" then the generated sequence is named "chart_data_id_seq" which is much nicer to read.

dev/openchpl_data-model.sql
```diff
@@ -27,6 +27,24 @@ CREATE TYPE openchpl.attestation as enum('Affirmative', 'Negative', 'N/A');
 CREATE TYPE openchpl.validation_message_type as enum('Error', 'Warning');
 CREATE TYPE openchpl.job_status_type as enum('In Progress', 'Complete', 'Error');
 
+CREATE TABLE openchpl.chart_data(
```
kekey1: I'm not sure if I'm being too nitpicky here, but putting the word "chart" in the table is tying this data with the way it's being presented for this task. While we aren't using it for anything else right now, we could in the future. If we wanted to do something with it in Excel or PDF then it's may not be specific to charts anymore. I don't have a specific suggestion for what it should be called, just asking you to think about it a little more. Maybe something like statistics and statistics_type.

andlar: 👍

PR #110, opened 2017-10-04
OCD-1789 - update allowable 2015 'Measures Succesfully Tested for G1/G2'

brianlindsy:


v-next.sql
```diff
@@ -0,0 +1,90 @@
...
+INSERT INTO openchpl.macra_criteria_map (criteria_id, value, name, description, last_modified_user)
+SELECT '170.315 (a)(1)',
+	'GAP-EH/CAH',
+	'(Gap Certified) Computerized Provider Order Entry - Medications: Eligible Hospital/Critical Access Hospital', 
+   'Required Test 10: Stage 2 Objective 3 Measure 1 and Stage 3 Objective 4 Measure 1', 
+   -1
+WHERE NOT EXISTS (
+    SELECT * 
+	FROM openchpl.macra_criteria_map (criteria_id, value, name, description, last_modified_user)
+	WHERE  ceriteria_id = '170.315 (a)(1)',
```
kekey1: this looks like a typo on "ceriteria_id", but I also think this value is not supposed to be the human-readable string that is the criteria number. I think it's supposed to be a lookup of an ID from another table 

(SELECT certification_criterion_id from openchpl.certification_criterion where number = '170.315 (a)(10)')

PR #114, opened 2017-10-13
OCD-1572: Mark old 2015 as deprecated

andlar:


andlar:
I thought of that, but it was easier to just copy and paste. If I did do the insert I'd have had to insert "false" for each of the other ones, when they were defaulted to false...

PR #117, opened 2017-10-24
Ocd-1803

brianlindsy:


v-next.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO openchpl.upload_template_version (name, available_as_of_date, header_csv, last_modified_user)
```
andlar: If we run v-next.sql more than once it'll add the new version multiple times. I think the idiom we use is something like "insert <blah> where not exists (select something from that table)" though I'm not positive. Suggest checking previous vM.N.O.sql files for samples

PR #118, opened 2017-10-27
Ocd 1827

kekey1:


v-next.sql
```diff
@@ -1,8 +1,209 @@
...
+DROP TABLE IF EXISTS openchpl.questionable_activity_version;
+DROP TABLE IF EXISTS openchpl.questionable_activity_product;
+DROP TABLE IF EXISTS openchpl.questionable_activity_developer;
+DROP TABLE IF EXISTS openchpl.questionable_activity_listing;
+DROP TABLE IF EXISTS openchpl.questionable_activity_certification_result;
+DROP TABLE IF EXISTS openchpl.questionable_activity_trigger;
+DROP TYPE IF EXISTS openchpl.questionable_activity_trigger_level;
+
+--sort of an enum of the possible objects that can trigger activity
+CREATE TYPE openchpl.questionable_activity_trigger_level as enum('Version', 'Product', 'Developer', 'Listing', 'Certification Criteria');
```
andlar: Will this be created multiple times?

kekey1: The line above your comment drops the enum and the line you commented on re-creates it. So it's only in there once!

andlar: Darn it! I had even looked up at that section but somehow missed the `TYPE` :)

PR #125, opened 2017-11-17
Ocd 1832 - QA history script

brianlindsy:


dev/fillInQAHistory.sql
```diff
@@ -0,0 +1,193 @@
+/* 1 - Certification Criteria Added */
+INSERT INTO openchpl.questionable_activity_listing (questionable_activity_trigger_id, listing_id, before_data, after_data, activity_date, activity_user_id, last_modified_user, creation_date, last_modified_date)
+(SELECT (SELECT id FROM openchpl.questionable_activity_trigger WHERE name = 'Certification Criteria Added'), listing_id, null, added_certs, creation_date, activity_user_id, last_modified_user, creation_date, creation_date FROM
```
kekey1: This is more of question, but I was curious about what the SELECT part returned here so I ran it locally and got one result. I have a fairly recent copy of the database (two-ish weeks old) and so also looked at questionable activity on the dev system to see if there were more criteria added than 1 since it seemed like there probably should be. There are and I guess I'm just wondering if either I ran this incorrectly to only get 1 result or if it's not returning everything; if the latter, did you compare the output of these queries to the questionable activity that is presented in the UI? I would think just do the comparison for a set period of time to try to get a match.

dev/fillInQAHistory.sql
```diff
@@ -0,0 +1,193 @@
...
+THEN (SELECT string_agg(number || ':' || g2Success, ';') FROM ((SELECT * FROM json_to_recordset(original_data::json->'certificationResults') as results(number text, g2Success text)) EXCEPT ALL (SELECT * FROM json_to_recordset(new_data::json->'certificationResults') as results(number text, g2Success text))) diff) END as new, creation_date, last_modified_user as activity_user_id, last_modified_user 
+FROM openchpl.activity WHERE activity_object_concept_id = (SELECT activity_concept_id FROM openchpl.activity_concept WHERE concept = 'CERTIFIED_PRODUCT')) as changed_name
+WHERE new IS NOT NULL)
+
+/* TODO: 7 - Measure Successfully Tested for 170.315 (g)(1) Added */
+INSERT INTO openchpl.questionable_activity_certification_result ((SELECT id FROM openchpl.questionable_activity_trigger WHERE name = 'Measure Successfully Tested for 170.315 (g)(1) Added'), certification_result_id, null, "170.315 (b)(3):EC Individual (TIN/NPI)", 2017-09-10 18:52:30.569, 5, 5, 2017-09-10 18:52:30.569, 2017-09-10 18:52:30.569)
```
kekey1: Does this work? I feel like the syntax is not right. Where is certification_result_id defined or selected? Do the dates need to be in quotes? Didn't we find more than just 1 instance of g1 macra measures added on prod/stg?

dev/fillInQAHistory.sql
```diff
@@ -0,0 +1,193 @@
...
+
+/* TODO: 10 - Measure Successfully Tested for 170.315 (g)(2) Removed */
+INSERT INTO openchpl.questionable_activity_certification_result ((SELECT id FROM openchpl.questionable_activity_trigger WHERE name = 'Measure Successfully Tested for 170.315 (g)(2) Removed'), certification_result_id, "170.315 (a)(1):EH/CAH;170.315 (a)(10):EH/CAH;170.315 (a)(13):EH/CAH;170.315 (a)(2):EH/CAH;170.315 (a)(3):RT8 EH/CAH,RT7 EH/CAH;170.315 (b)(2):EH/CAH;170.315 (b)(3):EH/CAH;170.315 (e)(1):RT2a EH/CAH,RT4a EH/CAH;170.315 (e)(2):EH/CAH;170.315 (e)(3):EH/CAH", null, 2017-08-22 21:25:17.046, -2, -2, 2017-08-22 21:25:17.046, 2017-08-22 21:25:17.046)
```
kekey1: Similar comment to above.

PR #126, opened 2017-11-21
Ocd 1903 - consolidate acb and atl roles

brianlindsy:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,16 @@
 # Release Notes
 
+### Minor Changes
+* Change ROLE_ACB_ADMIN to ROLE_ACB
+* Change ROLE_ATL_ADMIN to ROLE_ATL
+* Remove ROLE_ACB_STAFF
+* Remove ROLE_ATL_STAFF
+
+### Data changes
+* Mark deprecated upload templates as deleted
```
andlar: This part is going to have to wait until January 1st; the ACBs won't be ready for it until then

dev/openchpl_preload.sql
```diff
@@ -1302,14 +1302,12 @@ INSERT INTO openchpl.muu_accurate_as_of_date (accurate_as_of_date, last_modified
 INSERT INTO user_permission (user_permission_id, "name", description, authority, last_modified_user) VALUES
 (-2, 'ADMIN', 'This permission confers administrative privileges to its owner.', 'ROLE_ADMIN', -1),
 (1, 'USER_CREATOR' ,'This permission allows a user to create other users', 'ROLE_USER_CREATOR' , -1),
-(2, 'ACB_ADMIN' ,'This permission gives a user write access to their ACBs.', 'ROLE_ACB_ADMIN' , -1),
-(3, 'ACB_STAFF' ,'This permission gives a user read access to their ACBs', 'ROLE_ACB_STAFF' , -1),
-(4, 'ATL_ADMIN' ,'This permission gives a user write access to their ATLs.', 'ROLE_ATL_ADMIN' , -1),
-(5, 'ATL_STAFF' ,'This permission gives a user write access to their ATLs.', 'ROLE_ATL_STAFF' , -1),
+(2, 'ACB' ,'This permission gives a user write access to their ACBs.', 'ROLE_ACB_ADMIN' , -1),
+(4, 'ATL' ,'This permission gives a user write access to their ATLs.', 'ROLE_ATL_ADMIN' , -1),
```
andlar: Shouldn't the last item be `ROLE_ACB` and `ROLE_ATL`? Without the `_ADMIN` parts?

dev/openchpl_preload.sql
```diff
@@ -1302,14 +1302,12 @@ INSERT INTO openchpl.muu_accurate_as_of_date (accurate_as_of_date, last_modified
...
 (6, 'CMS_STAFF' ,'This permission gives a user read access to CMS reports.', 'ROLE_CMS_STAFF' , -1),
 (7, 'ONC_STAFF' ,'This permission gives a user access to the CMS Download file and report navigation section. It denies editing of Users/Products/CPs/ACBs/etc. No user invitation ability.', 'ROLE_ONC_STAFF' , -1);
 
-SELECT pg_catalog.setval('user_permission_user_permission_id_seq', 8, true);
+SELECT pg_catalog.setval('user_permission_user_permission_id_seq', 6, true);
```
andlar: Setting this to 6 means that the next roles that are added would overwrite the `CMS_STAFF` and `ONC_STAFF` roles

ocd-1903.sql
```diff
@@ -0,0 +1,38 @@
+UPDATE openchpl.global_user_permission_map as gupm
+SET user_permission_id_user_permission = 2
+WHERE user_id = 22;
+
+DELETE FROM openchpl.global_user_permission_map as gupm
+WHERE user_permission_id_user_permission = 3;
```
andlar: Instead of deleting them, shouldn't they be updated to be the permissions 2 or 4, from 3/5?

ocd-1903.sql
```diff
@@ -0,0 +1,38 @@
+UPDATE openchpl.global_user_permission_map as gupm
+SET user_permission_id_user_permission = 2
+WHERE user_id = 22;
+
```
kekey1: So if someone created an ACB_STAFF or ATL_STAFF user between now and the deployment of this (which I realize is unlikely), those new users wouldn't be included in this update. Would it be better to "SET user_permission_id_user_permission = 2 WHERE user_permission_id_user_permission = 3" and a second similar statement for ATL_STAFF?

ocd-1903.sql
kekey1: Agree with this. You should make an ACB_STAFF or ATL_STAFF user of your own, run this script, and make sure you can still log in with that user account. I don't think you'd be able to in this case because this would have deleted their permissions.

ocd-1903.sql
```diff
@@ -0,0 +1,61 @@
...
+UPDATE openchpl.global_user_permission_map as gupm
+SET user_permission_id_user_permission = 4
+WHERE user_permission_id_user_permission = 5
+AND NOT EXISTS (SELECT user_id FROM openchpl.global_user_permission_map WHERE (user_id, 4) IN (SELECT user_id, user_permission_id_user_permission FROM openchpl.global_user_permission_map));
+
+UPDATE openchpl.global_user_permission_map as gupm
+SET user_permission_id_user_permission = 2
+WHERE user_permission_id_user_permission = 3;
+
```
andlar: It looks like lines 11-13 will do the updates that lines 1-4 did, but without the check that there aren't duplicates. Is that right?

brianlindsy: Yes for some reason 1-4 did not update users permissions when they only have 3 as their permission.

brianlindsy:
I ran the latest sql script and there were errors about primary key constraints but the table looked updated properly.

PR #131, opened 2017-11-30
development - update v-next remove ocd-1903 script

brianlindsy:


brianlindsy:
Is that how the release notes should look when there's something on dev that wont go out and something on stg that will be in the next release?

andlar:
I don't think I'd put the planned release date into the STG section, but it doesn't hurt. It's possible we don't release on Monday (for whatever reason) and that would then be wrong, but it's not a big enough deal to be a problem

PR #141, opened 2017-12-21
OCD-1955: Soft-delete for Test Participants

andlar:


v-next.sql
```diff
@@ -120,7 +120,27 @@ BEGIN
...
+            ON (src.test_participant_id = ttpm.test_participant_id)
+        GROUP BY
+            src.test_participant_id
+        HAVING count(ttpm.test_participant_id) = 0
+    );
+    RETURN NEW;
+END;
+$$ language 'plpgsql';
+DROP TRIGGER IF EXISTS test_task_participant_map_soft_delete on openchpl.test_task_participant_map;
+CREATE TRIGGER test_task_participant_map_soft_delete AFTER UPDATE of deleted of deleted on openchpl.test_task_participant_map FOR EACH ROW EXECUTE PROCEDURE openchpl.test_task_participant_map_soft_delete();
```
kekey1: two "of deleted"s?

v-next.sql
```diff
@@ -120,7 +120,27 @@ BEGIN
 END;
 $$ language 'plpgsql';
 DROP TRIGGER IF EXISTS test_task_soft_delete on openchpl.test_task;
-CREATE TRIGGER test_task_soft_delete AFTER UPDATE on openchpl.test_task FOR EACH ROW EXECUTE PROCEDURE openchpl.test_task_soft_delete();
+CREATE TRIGGER test_task_soft_delete AFTER UPDATE of deleted on openchpl.test_task FOR EACH ROW EXECUTE PROCEDURE openchpl.test_task_soft_delete();
+
+CREATE OR REPLACE FUNCTION openchpl.test_task_participant_map_soft_delete()
+RETURNS TRIGGER AS $$
+BEGIN
+    UPDATE openchpl.test_participant as src SET deleted = NEW.deleted WHERE
```
kekey1: I changed this to 
`UPDATE openchpl.test_participant
    SET deleted = NEW.deleted 
    WHERE test_participant_id IN
    (
	select distinct tp.test_participant_id
	from openchpl.test_participant tp
	where not exists (
		select 1
		from openchpl.test_task_participant_map ttpm
		where tp.test_participant_id = ttpm.test_participant_id
		and ttpm.deleted = false
	)
    );`

which I think is a teeny bit faster. 

Also as you noted updating a listing via the API became terribly slow. All of the triggers were getting hit because the hibernate-generated update sql included the "deleted" field (turn on sql logging in log4j2.xml and the persistence.xml to see this). I don't think it SHOULD include the deleted field because that's not something we are letting anyone change via the API or UI, so I marked that column as not updateable. 

`@Column(nullable = false, updatable = false)
    protected Boolean deleted;`

in CertifiedProductEntity.java . Since the deleted flag is then not "updated" (even to the same value) when editing a listing, the performance is still good. 

With this in place, I also deleted a single test task with 37 participants from a listing and this seemed to work with the trigger, not too slow. If you are agreeable to these changes, that's probably what I think is the best at this point? Otherwise, another option is to drop the triggers at the end of this v-next and spend more time adding them back in during a later dedicated task.

PR #145, opened 2018-01-12
Ocd 1862

brianlindsy:


v-dev.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE openchpl.pending_certified_product 
+DROP COLUMN IF EXISTS has_qms
+ADD COLUMN has_qms boolean;
```
kekey1: I don't believe this file will actually run (did you test it??) I copied it locally and it did not work due to syntax errors. Something like

ALTER TABLE openchpl.pending_certified_product
DROP COLUMN IF EXISTS has_qms;

ALTER TABLE openchpl.pending_certified_product
ADD COLUMN has_qms boolean;

Would work.

PR #146, opened 2018-01-16
OCD-1983: Added new edition of CQMs

andlar:


andlar:
Verified it did load correctly on CI (at least as far as the reset-test-only / preload script is concerned)

PR #172, opened 2018-04-13
OCD-1897: Mark pending Listings as deleted if they have no ATL

andlar:


ocd-1897.sql
```diff
@@ -1,6 +1,12 @@
 --
 --OCD-1897 Cleanup
---  * removing old ATL column from certified_product table
+-- * removing old ATL column from certified_product table
+-- * as well as pending_certified_product table
+--   * do we actually want to do this? 
 --alter table openchpl.certified_product
 --drop column if exists testing_lab_id;
+--alter table openchpl.pendingcertified_product
```
tmy1313: Not a biggie, but the view is mispelled.

andlar: Valid. I put this file in place so we remember to do this clean up later. Will need to make sure it actually runs, and I'm not positive we actually want to do it, so this is more of just a placeholder for now

PR #198, opened 2018-06-07
Ocd 1503 -  Added column practice_type_id to test_functionality table

tmy1313:


PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,11 @@
+
+## Version m.n.o
```
andlar: The "m.n.o" in our documentation is a placeholder for actually filling in the version number. This file will eventually be used in deployment, and the idea is to have the expected version in this file, based on the current RELEASE_NOTES and PENDING_RELEASE_NOTES file versions

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,11 @@
...
+_Date TBD_
+
+### Table changes
+* Add practice type to the test functionality table
+
+### Data modifications
+* Updated 2014 functionality test records to specify their associated practice type
+
+---
```
andlar: I think there needs to be a blank line after line 11. Otherwise, during deployment, the `---` will butt up against the "previous" release notes section

dev/openchpl_data-model.sql
```diff
@@ -588,11 +588,15 @@ CREATE TABLE openchpl.test_functionality (
 	number varchar(200) not null,
 	name varchar(1000),
 	certification_edition_id bigint NOT NULL,
+	practice_type_id bigint NULL,
```
andlar: Should this be `practice_type_id bigint default null,`?

It might be valid, but looks odd. The way I usually test changes to these files is running the `reset-test-only.sh` script, and making sure no errors appear

v-next.sql
```diff
@@ -0,0 +1,36 @@
...
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 18;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 19;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 20;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 21;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 22;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 23;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 24;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 25;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 26;
+UPDATE openchpl.test_functionality SET practice_type_id = 2 WHERE test_functionality_id = 27;
```
andlar: I count eleven times in this file where a test_functionality is set to 2, but only ten in the preload file. Don't know which is right, but they can't both be

PR #200, opened 2018-06-13
Ocd 1503 backout

tmy1313:


PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -1,15 +0,0 @@
-
-## Version 14.3.0
-_Date TBD_
```
kekey1: Some of this stuff should still be here right? Quartz at least?

PR #212, opened 2018-08-28
Ocd 794 - User/Contact name changes

andlar:


ocd-794-cleanup.sql
```diff
@@ -0,0 +1,2 @@
+alter table openchpl.contact drop column if exists first_name;
+alter table openchpl.contact drop column if exists last_name;
```
kekey1: When should this file get run? Is it up to the prod deployer to remember or is it automated?

kekey1:
I think there is something in Bamboo that uses the contact table FYI... where we insert our own users? Make sure you change that bit of script when you deploy!

ocd-794-cleanup.sql
andlar: This file is a placeholder for the eventual part that should drop the old columns. I can't run it in the v-dev script because once the columns are dropped they can't be used again. There's another similar file in the root of this project for something we did a while ago. My plan is to leave them sitting around until we're sure everything is working, then put them into one of the upcoming releases.

PR #213, opened 2018-08-28
OCD-2414 - Update UI display values for a few G1G2 Macra Measures and Add a couple additional input values

tmy1313:


v-dev.sql
```diff
@@ -3,6 +3,208 @@ ALTER TABLE openchpl.api_key DROP COLUMN IF EXISTS whitelisted;
...
+values
+(
+	(select certification_criterion_id from openchpl.certification_criterion where "number" = '170.315 (a)(10)'),
+	'RT13 EH/CAH Stage 3',
+	'Query of Prescription Drug Monitoring Program (PDMP): Eligible Hospital/Critical Access Hospital',
+	'Required Test 13: Stage 3',
+	-1
+);
+
+insert into openchpl.macra_criteria_map
```
andlar: Why are all of these insert statements separate statements? Why not have them all as a single insert?

Also, what happens if this script runs twice? Will all of these inserts happen again?

tmy1313: Combined the inserts into a single statement.

If  the script were to run twice, it would violate a unique constraint and the data would not be inserted.
```
ERROR:  duplicate key value violates unique constraint "macra_criteria_unique"
DETAIL:  Key (criteria_id, value)=(10, RT13 EH/CAH Stage 3) already exists.
SQL state: 23505
```

v-dev.sql
```diff
@@ -3,6 +3,188 @@ ALTER TABLE openchpl.api_key DROP COLUMN IF EXISTS whitelisted;
 ALTER TABLE openchpl.api_key ADD COLUMN whitelisted boolean DEFAULT false;
 UPDATE openchpl.api_key SET whitelisted = true WHERE api_key_id = 1;
 
+-- OCD-2414 - Update UI display values for a few G1G2 Macra Measures and Add a couple additional input values
+insert into openchpl.macra_criteria_map
```
andlar: If this script is run twice, will it insert 2 copies of these items? If run a third time, will it insert three?

tmy1313: If the script were to run twice, it would violate a unique constraint and the data would not be inserted.
```
ERROR:  duplicate key value violates unique constraint "macra_criteria_unique"
DETAIL:  Key (criteria_id, value)=(10, RT13 EH/CAH Stage 3) already exists.
SQL state: 23505
```

PR #218, opened 2018-09-20
Modify api_key table to support the key deletion process

tmy1313:
OCD-2377


PENDING_RELEASE_NOTES.md
```diff
@@ -2,6 +2,12 @@
 ## Version 14.7.1
 _Date TBD_
 
+### Table Modifications
```
andlar: Should be in PENDING_PENDING.

v-next.sql
```diff
@@ -1258,3 +1258,20 @@ insert into openchpl.certification_result_test_tool (certification_result_id, te
...
     ((select cr.certification_result_id from openchpl.certification_result cr where cr.certified_product_id = 8300 and cr.certification_criterion_id = (select cc.certification_criterion_id from openchpl.certification_criterion cc where cc.number = '170.314 (c)(1)')), (select tt.test_tool_id from openchpl.test_tool tt where tt.name = 'Cypress' and tt.deleted = false), -1, '2.4.1'),
     ((select cr.certification_result_id from openchpl.certification_result cr where cr.certified_product_id = 8370 and cr.certification_criterion_id = (select cc.certification_criterion_id from openchpl.certification_criterion cc where cc.number = '170.314 (c)(1)')), (select tt.test_tool_id from openchpl.test_tool tt where tt.name = 'Cypress' and tt.deleted = false), -1, '2.4.1');
+
+-- OCD-2377 -  Cleanup of unused API keys
+ALTER TABLE openchpl.api_key ADD COLUMN last_used_date timestamp without time zone DEFAULT now();
```
andlar: Does this work multiple times? Does the function creation work multiple times?

Should also be in v-dev.sql

kekey1: This stuff should only be in v-dev and not v-next at all right? Due to STG being frozen?

PENDING_RELEASE_NOTES.md
tmy1313: addressed

PENDING_PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,10 @@
+
+## Version 14.8.0
+_Date TBD_
+
+### Table Modifications
+* API Key table
+  * Add column delete_warning_sent_date
+  * Add column last_used_date
+  * Add trigger to set delete_warning_sent_date = null when last_used_date is updated
+---
```
andlar: Should be an empty line both before and after the `---`

PENDING_RELEASE_NOTES.md
andlar: Sorry, I didn't follow that up with "...and not in PENDING

PR #225, opened 2018-11-14
2015 functionality tested should be restricted by criteria

tmy1313:
OCD-2532


dev/openchpl_data-model.sql
```diff
@@ -621,13 +621,13 @@ CREATE TABLE openchpl.test_functionality (
 	name varchar(1000),
 	certification_edition_id bigint NOT NULL,
 	practice_type_id bigint,
-	certification_criterion_id bigint,
+	certification_criterion_id_deleted bigint,
```
kekey1: I like this renaming

v-next.sql
```diff
@@ -669,3 +669,582 @@ AND NOT EXISTS (SELECT * FROM openchpl.cqm_criterion cc INNER JOIN openchpl.cqm_
...
+-- Rename the column that will no longer be used
+DO $$
+BEGIN
+  IF NOT EXISTS(SELECT *
+                FROM information_schema.columns
+                WHERE table_catalog = 'openchpl'
+                AND table_name = 'test_functionality'
+                AND column_name = 'certification_criterion_id')
+  THEN
+      ALTER TABLE "openchpl"."test_functionality" RENAME COLUMN "certification_criterion_id" TO "certification_criterion_id";
```
kekey1: Should you be renaming it to "_deleted" here and checking for the "_deleted" column's existence in the statement above?

v-next.sql
```diff
@@ -669,3 +669,582 @@ AND NOT EXISTS (SELECT * FROM openchpl.cqm_criterion cc INNER JOIN openchpl.cqm_
...
+    (SELECT test_functionality_id FROM openchpl.test_functionality WHERE number = '170.102(19)(ii)' AND  certification_edition_id = 3),
+	-1
+);
+
+-- Convert the existing 2014 data
+INSERT INTO openchpl.test_functionality_criteria_map (criteria_id, test_functionality_id, last_modified_user)
+VALUES (
+    (SELECT certification_criterion_id FROM openchpl.certification_criterion WHERE number = '170.314 (a)(6)'),
+    (SELECT test_functionality_id FROM openchpl.test_functionality WHERE number = '(a)(6)(i)' AND  certification_edition_id = 2),
+	-12
```
kekey1: Should be -1 I think

dev/openchpl_data-model.sql
andlar: Renaming it is fine in the `v-dev` or `v-next`, but in our `dev/data-model` file it should move directly to the final state. As long as that doesn't break our tests, I think our goal is to not have any of the OBE stuff in the `./dev/*` files

dev/openchpl_preload.sql
```diff
@@ -1146,7 +1146,7 @@ ITU-T E. 164, Series E: Overall Network Operation, Telephone Service, Service Op
 ('170.204(a)(2)','Web Content Accessibility Guidelines (WCAG) 2.0, Level AA Conformance', 3, -1);
 --2015
 
-INSERT INTO openchpl.test_functionality(number, name, certification_edition_id, practice_type_id, certification_criterion_id, last_modified_user) VALUES
+INSERT INTO openchpl.test_functionality(number, name, certification_edition_id, practice_type_id, certification_criterion_id_deleted, last_modified_user) VALUES
```
andlar: If we're not planning on using this column any more this should be removed from the preload entirely

v-next.sql
```diff
@@ -669,3 +669,582 @@ AND NOT EXISTS (SELECT * FROM openchpl.cqm_criterion cc INNER JOIN openchpl.cqm_
...
+(
+    (SELECT certification_criterion_id FROM openchpl.certification_criterion WHERE number = '170.314 (e)(1)'),
+    (SELECT test_functionality_id FROM openchpl.test_functionality WHERE number = '(e)(1)(i)(B)(1)(ii)' AND  certification_edition_id = 2),
+	-1
+);
+-- END OCD-2532
+----------------------------------------------------
+
+--re-run grants
+\i dev/openchpl_grant-all.sql
```
andlar: It's helpful to have a newline at the end of the file so we can cat things together, but not required

ocd-2532.sql
```diff
@@ -0,0 +1,2 @@
+--Remove the column test_functionality.certification_criterion_id_deleted
+ALTER TABLE openchpl.test_functionality DROP COLUMN certification_criterion_id_deleted;
```
andlar: Does this work if the column is already deleted?

Also should have newline at the end.

v-next.sql
```diff
@@ -669,3 +669,582 @@ AND NOT EXISTS (SELECT * FROM openchpl.cqm_criterion cc INNER JOIN openchpl.cqm_
...
+);
+
+--Create the required triggers
+CREATE TRIGGER test_functionality_criteria_map_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.test_functionality_criteria_map FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
+CREATE TRIGGER test_functionality_criteria_map_timestamp BEFORE UPDATE on openchpl.test_functionality_criteria_map FOR EACH ROW EXECUTE PROCEDURE openchpl.update_last_modified_date_column();
+
+-- Rename the column that will no longer be used
+DO $$
+BEGIN
+  IF NOT EXISTS(SELECT *
```
andlar: Shouldn't this be `If EXISTS(SELECT *`?

v-next.sql
```diff
@@ -669,3 +669,582 @@ AND NOT EXISTS (SELECT * FROM openchpl.cqm_criterion cc INNER JOIN openchpl.cqm_
 
 -- END OCD-2533
 ----------------------------------------------------
+
+-- OCD-2532 - 2015 functionality tested should be restricted by criteria
+-- 1) Create new table to support relationship between criteria and test functionality
+-- 2) Add triggers to new table
+-- 3) Rename old column
```
andlar: This shouldn't be here either

tmy1313: Just shoot me...  Made the change and didn't save it.

PR #226, opened 2018-11-15
Ocd 2523 - password reset

brianlindsy:


dev/openchpl_data-model.sql
```diff
@@ -53,6 +53,17 @@ CREATE TABLE openchpl.user(
 -- ALTER TABLE openchpl.user OWNER TO openchpl;
 -- ddl-end --
 
+CREATE TABLE openchpl.user_reset_token(
+	user_reset_token_id bigserial NOT NULL,
+	user_reset_token varchar(15) NOT NULL,
+	user_id bigint,
```
kekey1: Can the user id be null? I think it should have a foreign key constraint.

v-dev.sql
```diff
@@ -0,0 +1,13 @@
+DROP TABLE IF EXISTS openchpl.user_reset_token;
+CREATE TABLE openchpl.user_reset_token(
```
kekey1: Need to add the triggers in here as well.

PR #227, opened 2018-11-16
Add reason column to questionable_activity_developer table

tmy1313:
OCD-2560

### DIFF
--- a/v-dev.sql
+++ b/v-dev.sql
@@ -574,5 +574,9 @@ VALUES (
 -- END OCD-2532
 ----------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


v-dev.sql
```diff
@@ -574,5 +574,9 @@ VALUES (
 -- END OCD-2532
 ----------------------------------------------------
 
+-- OCD-2560 - Add Reason to Questionable Activity when changing Developer status
+-- Add column reason to questionable_activity_developer table
+ALTER TABLE openchpl.questionable_activity_developer ADD COLUMN reason TEXT DEFAULT null;
```
andlar: Does this work if it runs twice?

PR #231, opened 2018-11-27
Ocd 1665 - Allow ACBs and ATLs to be retired

kekey1:


kekey1:
Not merging this until after STG is frozen.

PR #233, opened 2018-12-04
Ocd 2554 Remove all notification* tables

kekey1:


v-dev.sql
```diff
@@ -72,4 +72,15 @@ UPDATE openchpl.testing_lab
...
 -- END OCD-1665 Retire ACBs and ATLS
+----------------------------------------------------
+
+----------------------------------------------------
+-- OCD-2554: Remove old notification infrastructure
+----------------------------------------------------
+DROP TABLE IF EXISTS openchpl.notification_type_recipient_map;
+DROP TABLE IF EXISTS openchpl.notification_recipient;
+DROP TABLE IF EXISTS openchpl.notification_type_permission;
+DROP TABLE IF EXISTS openchpl.notification_type;
```
andlar: Does it need a cascade or anything to get rid of the triggers?

kekey1: Nope! They go when the table gets dropped.

PR #245, opened 2019-01-16
Ocd 2655 - Allow ROLE_ONC, ROLE_ADMIN to manage pending surveillances

tmy1313:


dev/openchpl_data-model.sql
```diff
@@ -2370,6 +2370,7 @@ CREATE TABLE openchpl.pending_surveillance (
 	end_date date,
 	type_value varchar(30),
 	randomized_sites_used integer,
+	user_permission_id bigint NOT NULL,
```
kekey1: Can you give this a foreign key constraint like it has in the surveillance table?

PR #247, opened 2019-01-22
Ocd 2635 - update test task id and participant id to be 20 characters max

brianlindsy:


brianlindsy:
The alter will fail and say that the value is too long to update. I don't know anyway around that other than making sure all of the environments don't have values that are longer than 20 characters, which they shouldn't if the code checks are correct. Or manually changing the values before the update. I checked that against the latest prod backup. Ill check against the DEV and STG backup too.

PR #254, opened 2019-02-15
Ocd 2679 - CHPL managed user/ACB relationship

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -2,7 +2,12 @@
 ## Version 15.3.1
```
kekey1: This should probably be 15.4 now

ocd-2679.sql
```diff
@@ -0,0 +1,65 @@
...
+	deleted bool NOT NULL DEFAULT false,
+	CONSTRAINT user_certification_body_pk PRIMARY KEY (id),
+	CONSTRAINT user_fk FOREIGN KEY (user_id)
+		REFERENCES openchpl.user (user_id) 
+		MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT,
+	CONSTRAINT certification_body_fk FOREIGN KEY (certification_body_id)
+		REFERENCES openchpl.certification_body (certification_body_id) 
+		MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+);
+
```
kekey1: Needs the audit triggers added here and in the audit file.

ocd-2679.sql
```diff
@@ -0,0 +1,65 @@
...
+		on sid.id = entry.sid
+	inner join openchpl.acl_object_identity oi
+		on entry.acl_object_identity = oi.id
+	inner join openchpl.acl_class c
+		on oi.object_id_class = c.id
+where c.class = 'gov.healthit.chpl.dto.CertificationBodyDTO'
+and exists (select * from openchpl.user where user_id = sid.id);
+
+-- Backup acl_object_identity table
+CREATE TABLE openchpl.acl_object_identity_backup AS
```
kekey1: Do you need to drop these backup tables in order for the script to run multiple times?

ocd-2679.sql
```diff
@@ -0,0 +1,65 @@
...
+-- Backup acl_class table
+CREATE TABLE openchpl.acl_class_backup AS
+SELECT * FROM openchpl.acl_class;
+
+-- Backup acl_entry table
+CREATE TABLE openchpl.acl_entry_backup AS
+SELECT * FROM openchpl.acl_entry;
+
+-- Delete rows from acl_entry relacted to ACBs
+DELETE FROM openchpl.acl_entry
```
kekey1: If the script runs multiple times you probably can't drop the backup tables and also delete the ACB data from the live tables or we'll lose the data. This seems a candidate for our two-part process so this release would drop/create/populate the backup tables and then the next release would delete the now unused ACB data from the spring tables. There is some pre-loaded ACL data you will also want to remove in that following release.

PENDING_RELEASE_NOTES.md
tmy1313: Updated

ocd-2679.sql
tmy1313: updated in ocd-2679.sql and openchpl_audit.sql

PENDING_RELEASE_NOTES.md
```diff
@@ -1,8 +1,13 @@
 
-## Version 15.3.1
+## Version 15.4
 _Date TBD_
 
+### Data Modifications
```
andlar: This isn't a Data modification
```suggestion
### Table Modifications
```

PENDING_RELEASE_NOTES.md
```diff
@@ -1,8 +1,13 @@
...
+* Add table user_certification_body_map to store relationship between users and ACBs
+
 ### Data Modifications
 * Change Title & Descriptions for some G1/G2 Macra measures
+* Populate user_certification_body_map based on ACL tables
+* Backup ACL tables and delete ACB related information from them
```
andlar: Are we deleting on this build? Or has that moved to the next push?

ocd-2679.sql
andlar: I think part of the idea was to remove the data from the ACL tables to make sure there wasn't anything left over from the spring stuff that was still using it, and that's why the backup tables are being created. That said, the multiple run thing is definitely an issue. Maybe this feature will need to be a three step, where the first one doesn't delete the data from the ACL table, but just copies it to the backup, the second deletes the data, and the third deletes the backup tables

PENDING_RELEASE_NOTES.md
```diff
@@ -1,8 +1,14 @@
 
-## Version 15.3.1
+## Version 15.4
```
andlar:
```suggestion
## Version 15.4.0
```

PENDING_RELEASE_NOTES.md
```diff
@@ -1,8 +1,14 @@
...
 _Date TBD_
 
+### Table Modifications
+* Add table user_certification_body_map to store relationship between users and ACBs
+* Create temporary backup tables for ACL data
+
 ### Data Modifications
 * Change Title & Descriptions for some G1/G2 Macra measures
+* Populate user_certification_body_map based on ACL tables
+* Backup ACL tables and delete that have ACB related information
```
andlar: The "delete" part isn't happening on this push, right? 

I don't think you need to include data modifications when those tables are being created in the same build. I assume tables that are created get populated during that process, though if you want to make it explicit I'd put that in the "table modification" section. E.g., 
```
* Add table...and ACBs. Populate table at that time
```

dev/openchpl_data-model.sql
```diff
@@ -2982,6 +2982,24 @@ CREATE TABLE openchpl.chpl_file
 		ON DELETE RESTRICT
 );
 
+CREATE TABLE IF NOT EXISTS openchpl.user_certification_body_map (
+	id bigserial NOT NULL,
+	user_id bigint NOT NULL,
+	certification_body_id bigint NOT NULL,
+	retired bool NOT NULL DEFAULT false,
```
andlar: What is retired? Users can't be retired from an ACB, right?

kekey1: I asked this question on the API PR and Todd said that the relationship between user and ACB gets retired. I thought that was something new but just figured I missed it whenever we put that in.

andlar: It does? When? Why?

ocd-2679.sql
```diff
@@ -0,0 +1,125 @@
...
+where c.class = 'gov.healthit.chpl.dto.CertificationBodyDTO'
+and exists (select * from openchpl.user where user_id = sid.id);
+
+--Add audit triggers
+CREATE TRIGGER user_certification_body_map_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.user_certification_body_map FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
+CREATE TRIGGER user_certification_body_map_timestamp BEFORE UPDATE on openchpl.user_certification_body_map FOR EACH ROW EXECUTE PROCEDURE openchpl.update_last_modified_date_column();
+
+
+DO $$
+BEGIN
```
andlar: This needs to be done before the mapping table is created, right? Otherwise, on a second run, there won't be any ACB related ACL entries to add to the mapping table

tmy1313: Good point...  Addressed.

ocd-2679.sql
```diff
@@ -0,0 +1,128 @@
...
+		on oi.object_id_class = c.id
+where c.class = 'gov.healthit.chpl.dto.CertificationBodyDTO'
+and exists (select * from openchpl.user where user_id = sid.id);
+
+--Add audit triggers
+CREATE TRIGGER user_certification_body_map_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.user_certification_body_map FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
+CREATE TRIGGER user_certification_body_map_timestamp BEFORE UPDATE on openchpl.user_certification_body_map FOR EACH ROW EXECUTE PROCEDURE openchpl.update_last_modified_date_column();
+
+
+DO $$
```
andlar: Are these lines still supposed to be here? Have you run this a few times and made sure it worked as expected?

tmy1313: The DO $$ can be refactored out...   Addressed.

ocd-2679.sql
```diff
@@ -0,0 +1,119 @@
...
+  -- Backup acl_class table
+  DROP TABLE IF EXISTS openchpl.acl_class_backup;
+
+  CREATE TABLE openchpl.acl_class_backup AS
+  SELECT * FROM openchpl.acl_class;
+
+  DELETE FROM openchpl.acl_class
+  WHERE "class" = 'gov.healthit.chpl.dto.CertificationBodyDTO';
+
+END $$;
```
andlar: Is this supposed to be here? Does it give an error when you run the script?

tmy1313: Yes, it is supposed to be there. It is the closing for 
```
DO $$
BEGIN
```
I have run the script several times, without issues.

andlar: Why is it here? We don't usually have our SQL stuff bracketed with this

tmy1313: It is necessary to the `IF ... THEN` to run correctly.  I tried removing it, and I get the following error:
```
ERROR:  syntax error at or near "IF"
LINE 3:   IF EXISTS(SELECT * FROM information_schema.tables WHERE ta...
          ^
SQL state: 42601
Character: 116
```

andlar:
My understanding of the DO blocks is that they're used for anonymous
functions, where things like the IF can be used. I think I'd prefer that
those anonymous functions be scoped to as small a set of code as possible.
When the function has all of the other SQL statements, do those product
output when you run the script? If not that's definitely an issue

ocd-2679.sql
andlar: My understanding of the DO blocks is that they're used for anonymous functions, where things like the IF can be used. I think I'd prefer that those anonymous functions be scoped to as small a set of code as possible. When the function has all of the other SQL statements, do those product output when you run the script? If not that's definitely an issue

tmy1313: I'll wrap each IF in a DO block, for scoping purposes.  I wasn't really looking at the output -- when there is an error inside a DO, it outputs the error to the console.  When it completes successfully, there is no usable output.

andlar: I'd be okay with all of the IFs in a single block, but I do think we need to see the other SQL update outputs as they run

andlar: Thanks!

v-next.sql
```diff
@@ -4,6 +4,7 @@ insert into openchpl.data_model_version (version, deploy_date, last_modified_use
 
 \i ocd-2697.sql
 \i ocd-2633.sql
+\i ocd-2697.sql
```
andlar: Finally figured out why my stuff wasn't working. It's `97` vs `79` again

PR #257, opened 2019-02-19
Ocd 2697 - friendlier upload handling

andlar:


dev/openchpl_data-model.sql
```diff
@@ -1689,9 +1689,9 @@ CREATE TABLE openchpl.pending_test_participant (
 	education_type_id bigint,
 	user_entered_education_type varchar(250),
 	occupation varchar(250),
-	professional_experience_months int,
-	computer_experience_months int,
-	product_experience_months int,
+	professional_experience_months varchar(25),
```
kekey1: What happens if they have something longer than 25 characters? Want to just avoid all error possibilities and make it a "text" column?

andlar: ~Then they're being dumb?~ It's supposed to be a number. I figured 25 characters was plenty of space to put any reasonable number in.

I was going to say something about using varchar for performance reasons, but according to the docs there's no performance benefit, so, as much as it pains me to support that behavior I'll change it around

kekey1: Haha I agree, but was just thinking they didn't put in numbers to begin with which is why we're making the change so I didn't think we should assume they'll put in only non-numbers of 25 characters or less! Thanks for checking on performance and updating.

PR #270, opened 2019-04-10
Ocd 2817 - Uploading an MUU file twice, with the same effective date, seems to make some listings disappear

tmy1313:


dev/openchpl_views.sql
```diff
@@ -529,7 +529,7 @@ LEFT JOIN (
 		muu.meaningful_use_users_date
 	FROM openchpl.meaningful_use_user muu
 	 INNER JOIN ( 
-	   SELECT meaningful_use_user.certified_product_id,
+	   SELECT DISTINICT meaningful_use_user.certified_product_id,
```
andlar:
```suggestion
	   SELECT DISTINCT meaningful_use_user.certified_product_id,
```

andlar:
It certainly wouldn't hurt to look

PR #275, opened 2019-04-22
OCD-2842: Improve banned developer page

kekey1:


kekey1:
Yes... in the other places it was included in one of the joins but not SELECTed anywhere else in the view so I just removed it since it wasn't being used.

PR #279, opened 2019-05-08
Ocd 2885 (OCD-2820 - Resolve phantom logged in issue)

tmy1313:


kekey1:
Oh also please pull the latest development - pending release notes should not have anything else in them besides this ticket.

ocd-2885.sql
```diff
@@ -0,0 +1,55 @@
+DROP TABLE IF EXISTS openchpl.ff4j_features;
```
andlar: Don't these need to be reversed so they can drop correctly?

kekey1:
> Do we want to / can we / put these in a different schema than the regular CHPL data?

I had the same comment on the API ticket - so I think maybe we should!

dev/reset-test-only.sh
```diff
@@ -10,6 +10,7 @@ if [ $# -ne 2 ]; then
     psql -Upostgres -f openchpl_preload.sql openchpl_test
     psql -Upostgres -f openchpl_api-key.sql openchpl_test
     psql -Upostgres -f openchpl_quartz.sql openchpl_test
+	psql -Upostgres -f openchpl_ff4j.sql openchpl_test
```
andlar: It's tiny, but these various "different indentation" things irk the heck out of me

ocd-2885.sql
```diff
@@ -0,0 +1,56 @@
+DROP SCHEMA IF EXISTS ff4j CASCADE;
+
+CREATE SCHEMA ff4j
```
andlar: Why not just do the `\i dev/ff4j-schema.sql` file thing? This looks like that same code, or at least SUPER-CLOSE

ocd-2885.sql
```diff
@@ -0,0 +1,3 @@
+DROP SCHEMA IF EXISTS ff4j CASCADE;
+
+\i dev/ff4j-schema.sql
```
kekey1: Just tried running my v-next locally to get the schema change but the filename you have here is not the file that has the schema. Should be openchpl_ff4j.sql

PR #282, opened 2019-05-22
Ocd 2789 - Improve Reports responsiveness - page load - User Activity

tmy1313:


v-next.sql
```diff
@@ -1,5 +1,4 @@
-\i ocd-2885.sql
```
kekey1: 2885 is the FF4J stuff and I think we still need that here.

dev/openchpl_preload.sql
```diff
@@ -2281,4 +2281,6 @@ VALUES
 	('Developer Report', -1),
 	('Listing Report', -1),
 	('Product Report', -1),
-	('Version Report', -1);
+	('Version Report', -1),
+    ('User Report', -1),
+    ('User Action Report', -1);
```
andlar: Can we keep indentation consistent? Or at least not add new inconsistencies?

PR #288, opened 2019-06-18
Ocd 2913 - Associate a listing to a complaint

tmy1313:


v-next.sql
```diff
@@ -1,5 +1,4 @@
-\i ocd-2912.sql
-\i ocd-2916.sql
+\i ocd-2913.sql
 \i ocd-2917.sql
 
 \i dev/openchpl_soft-delete.sql
```
kekey1: In this soft delete file there are a bunch of triggers that sort of "cascade" setting deleted to true for all the things related to a listing (criteria, cqms, etc etc) if a listing is marked as deleted. Can you add something in there for complaint_listing_map? That's assuming if a listing is marked deleted then its mapping to any complaints should also be marked as deleted... which I think makes sense. If a complaint gets marked as deleted do you handle the cascading of that in the API code or want to do it in the soft delete file? I don't think we have been good about always making sure that related things are marked deleted so just trying to keep that in mind more.

PR #301, opened 2019-07-25
Ocd 3016 - Track activity for complaints

tmy1313:


PENDING_RELEASE_NOTES.md
```diff
@@ -0,0 +1,7 @@
+## Version 17.8.0
```
andlar: If it's just a data modification, that's a "bugfix" level of the version change. Updates to table structure / views fall into the minor category, and things that break compatibility with the API are major ones
```suggestion
## Version 17.7.1
```

PR #302, opened 2019-07-30
feat: Add privileged surveillance data

kekey1:
* Add lookup tables for surveillance_process_type and surveillance_outcome values
* Add mapping table for privileged data associated with surveillance and quarterly report

OCD-2920


PENDING_RELEASE_NOTES.md
```diff
@@ -1,7 +1,8 @@
-## Version 17.7.1
+## Version 17.9.0
```
andlar: Might need to merge from upstream/development again. I don't think there should be any PENDING notes on development, and probably never again, assuming the deployment scripts work

Wouldn't hurt to update the complete RELEASE NOTES too, if you can figure out what was actually released

PR #303, opened 2019-07-30
 fix: add activity concept for type COMPLAINT

tmy1313:
qa: reset activity concept sequence and add COMPLAINT, if necessary

OCD-3016

### DIFF
--- a/dev/openchpl_preload.sql
+++ b/dev/openchpl_preload.sql
@@ -1891,7 +1891,7 @@ INSERT INTO openchpl.qms_standard(name, last_modified_user) VALUES
 ('ISO 13485', -1),
 ('IEC 62304', -1);
 
-INSERT INTO openchpl.activity_concept (activity_concept_id, concept, last_modified_user) VALUES
+INSERT INTO openchpl.activity_concept (concept, last_modified_user) VALUES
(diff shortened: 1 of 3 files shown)


dev/openchpl_preload.sql
```diff
@@ -1892,22 +1892,22 @@ INSERT INTO openchpl.qms_standard(name, last_modified_user) VALUES
 ('IEC 62304', -1);
 
 INSERT INTO openchpl.activity_concept (activity_concept_id, concept, last_modified_user) VALUES
```
kekey1: Don't think this change will work because the INSERT statement still lists the activity_concept_id column. Make sure to run reset-test-only.sh after changing this to confirm no errors.

ocd-3016.sql
```diff
@@ -0,0 +1,16 @@
+do $$
+begin
+    if exists(select count(*) from openchpl.activity_concept where concept = 'COMPLAINT') THEN
+        -- Just set the seq to the correct value
+        perform setval('openchpl.activity_concept_activity_concept_id_seq', 16);
```
kekey1: This seems like it will work..not necessarily requesting a change, but would it be simpler to set the sequence to the count(*) value of the activity_concept table?

PR #318, opened 2019-09-11
feat: Add tables and views to support reporting on bad urls

kekey1:
* Add url_check_result table to store all the results of all urls that have been checked
* Add url_type table as a lookup table of places urls can be found in the system
* Update certified_product_summary view to have developer contact info

OCD-2841


dev/openchpl_views.sql
```diff
@@ -1004,10 +1004,10 @@ WHERE deleted = false;
 
 CREATE VIEW openchpl.certified_product_summary AS
  SELECT cp.certified_product_id,
+	(select chpl_product_number from openchpl.get_chpl_product_number(cp.certified_product_id)),
```
andlar: I wonder if this is going to fix part of OCD-3017

PR #319, opened 2019-09-12
feat: Add tables to support persisting Website Change Requests

tmy1313:
OCD-2159


ocd-2159.sql
```diff
@@ -0,0 +1,158 @@
...
+	    REFERENCES openchpl.change_request_status_type (id)
+        MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT,
+    CONSTRAINT certification_body_fk FOREIGN KEY (certification_body_id)
+	    REFERENCES openchpl.certification_body (certification_body_id)
+        MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+);
+CREATE TRIGGER change_request_status_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.change_request_status FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
+CREATE TRIGGER change_request_status_timestamp BEFORE UPDATE on openchpl.change_request_status FOR EACH ROW EXECUTE PROCEDURE openchpl.update_last_modified_date_column();
+
+CREATE TABLE openchpl.change_request_certification_body_map (
```
kekey1: Just wondering about this table and how it's expected to be used... I went back and looked at the requirements in the Epic and see that the system should notify "Any ACB that the Developer has a Listing with" about change requests so I thought maybe for that? But the listing's ACB can be changed so I think we'd calculate "Any ACB that the developer has a listing with" on the fly in one of the permissions classes rather than save it here. Elsewhere in the tickets, the way things are written it's more like any ACB should be able to view/manage any request and in that case I don't think you'd need a mapping. So anyway, I just wanted to understand this table. I haven't dug through the 90+ API files changed yet to figure it out myself. :)

andlar: I think this is for saving the "acting ACB" when a change is modified, though I'm only looking at the comment, so might be thinking of a different table

kekey1: Should it be joined with the change_request_status table then?

kekey1: Should it be joined with the change_request_status table then?

kekey1: Or actually the certification body id would just be a column in that status table then right?

PR #321, opened 2019-09-30
feat: Add view to determine relationship between developers and acbs

tmy1313:
OCD-3093


ocd-3093.sql
```diff
@@ -0,0 +1,22 @@
+CREATE OR REPLACE VIEW openchpl.developer_certification_body_map
```
andlar: This isn't needed in this file. The v-next.sql file inserts the openchpl_views.sql file, so we don't need to put views (or soft-delete / grant-all statements) in ocd-nnnn files

v-next.sql
```diff
@@ -1,3 +1,5 @@
+\i ocd-2159.sql
```
andlar: Seeing this file injected here makes me wonder if the changes you made as a result of 3093 could be in the 2159 file. As in, I think the 3093 file could be removed completely if the new user_permission_id is just added to the table definition in 2159. The view is already in the openchpl_views file, so there won't be a need to even have that one. And 2159 hasn't been promoted or resolved or anything, so why make it more complicated than it needs to be?

PR #333, opened 2019-11-13
feat: add user_permission_id to the change_request_status table

tmy1313:
OCD-3093

### DIFF
--- a/dev/openchpl_data-model.sql
+++ b/dev/openchpl_data-model.sql
@@ -3187,6 +3187,7 @@ CREATE TABLE openchpl.change_request_status (
     status_change_date timestamp NOT NULL,
     comment text,
     certification_body_id bigint,
+    user_permission_id bigint NOT NULL,
     creation_date timestamp NOT NULL DEFAULT NOW(),
 	last_modified_date timestamp NOT NULL DEFAULT NOW(),
 	last_modified_user bigint NOT NULL,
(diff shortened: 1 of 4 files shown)


dev/openchpl_data-model.sql
```diff
@@ -3201,6 +3202,9 @@ CREATE TABLE openchpl.change_request_status (
     CONSTRAINT certification_body_fk FOREIGN KEY (certification_body_id)
 	    REFERENCES openchpl.certification_body (certification_body_id)
         MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+    CONSTRAINT user_permission_fk FOREIGN KEY (user_permission_id)
+	    REFERENCES openchpl.user_permission (user_permission_id)
+        MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT;
```
andlar: Is this semi-colon needed? There wasn't one before on the constraint just before it

andlar: Also, the previous constraint probably needs a comma at the end of line 3204

ocd-3093.sql
```diff
@@ -0,0 +1,67 @@
+DROP TABLE IF EXISTS openchpl.change_request_website;
```
andlar: These tables had triggers associated with them, right? Do these DROP commands drop those triggers too?

PR #352, opened 2020-02-28
OCD-3220: Join criterion id into pending surveillance and nonconformity statistics

kekey1:


ocd-3220.sql
```diff
@@ -0,0 +1,27 @@
...
+		REFERENCES openchpl.certification_criterion (certification_criterion_id)
+		MATCH FULL ON DELETE CASCADE ON UPDATE CASCADE;
+
+ALTER TABLE openchpl.nonconformity_type_statistics
+ADD COLUMN certification_criterion_id bigint;
+
+ALTER TABLE openchpl.nonconformity_type_statistics
+ADD CONSTRAINT certification_criterion_fk FOREIGN KEY (certification_criterion_id)
+		REFERENCES openchpl.certification_criterion (certification_criterion_id)
+		MATCH FULL ON DELETE CASCADE ON UPDATE CASCADE;
```
andlar: Could you put a newline on this line? When the deployment script runs it joins all of the ocd-nnnn files together, and having newlines at the end of each make the vM.N.O script look a little nicer

It's not required, because the script does assume people _might_ not have those newlines

I'll approve as written

dev/openchpl_views.sql
```diff
@@ -1152,3 +1153,20 @@ AS SELECT DISTINCT cp.certification_body_id,
...
+where sn.deleted = false
+and cc is null
+group by sn.nonconformity_type)
+union 
+(select count(cc.certification_criterion_id) as nonconformity_count, null as non_criterion_type, cc.certification_criterion_id as certification_criterion_id
+from openchpl.surveillance_nonconformity sn
+join openchpl.certification_criterion cc on cc.certification_criterion_id = sn.certification_criterion_id
+where sn.deleted = false and cc.removed = false
+group by cc.certification_criterion_id
+order by cc.certification_criterion_id)) as subquery;
```
tmy1313: Do we need an EOL here?

andlar: It's not needed for any reason like the one I asked for in the OCD-nnnn file. I'd prefer that we have it, but that's just a preference on my part; there's no functional need for it as we don't join this file with others at any time

PR #354, opened 2020-03-20
OCD-3252: Save "Cures Update" per Listing data

andlar:


dev/openchpl_data-model.sql
```diff
@@ -1412,6 +1412,20 @@ CREATE TABLE openchpl.certification_status_event (
 		MATCH FULL ON DELETE RESTRICT ON UPDATE CASCADE
 );
 
+CREATE TABLE openchpl.cures_update_event (
+        cures_update_event_id bigserial NOT NULL,
```
kekey1: Can we just use "id" here? The type bigserial means that postgres made a sequence called "cures_update_event_cures_update_event_id_seq" but if we just named it "id" then the sequence would be called "cures_update_event_id_seq".

ocd-3252.sql
```diff
@@ -0,0 +1,20 @@
...
+	creation_date timestamp NOT NULL DEFAULT NOW(),
+	last_modified_date timestamp NOT NULL DEFAULT NOW(),
+	last_modified_user bigint NOT NULL,
+	deleted bool NOT NULL DEFAULT false,
+	CONSTRAINT cures_update_event_pk PRIMARY KEY (cures_update_event_id),
+	CONSTRAINT certified_product_fk FOREIGN KEY (certified_product_id) REFERENCES openchpl.certified_product (certified_product_id)
+		MATCH FULL ON DELETE RESTRICT ON UPDATE CASCADE
+);
+
+\i dev/openchpl_views.sql
```
kekey1: does this line need to be both here and in the v-next?

andlar: It does, actually. Dropping the TABLE above also dropped the views, and I need the views back in order to use them to get the data I need to fill the new TABLE

PR #358, opened 2020-04-07
OCD-2162: Support change requests for all developer details

kekey1:


ocd-2162.sql
```diff
@@ -0,0 +1,35 @@
...
+	deleted bool NOT NULL DEFAULT false,
+	CONSTRAINT change_request_developer_details_pk PRIMARY KEY (id),
+    CONSTRAINT change_request_fk FOREIGN KEY (change_request_id)
+	    REFERENCES openchpl.change_request (id)
+        MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+);
+
+delete from openchpl.change_request_status
+where change_request_id in
+  (select id from openchpl.change_request where change_request_type_id = 2);
```
andlar, line 33: We might want this to be a sub-select looking for the same name as the type that was inserted in the first few lines of this file, though I think it would only matter if we were doing two types of new CRs at the same time

PR #360, opened 2020-04-30
OCD-3335: Add auditing triggers to quartz, ff4j, openchpl tables

kekey1:


ocd-3335.sql
```diff
@@ -0,0 +1,43 @@
+-- triggers missing from existing openchpl tables
+DROP TRIGGER IF EXISTS complainant_type_audit on openchpl.complainant_type;
```
andlar, line 2: I don't see these complianant triggers in the data-model.sql file. Should they be?

kekey1: Triggers for this table are already in the audit file that adds all the triggers for openchpl tables so I did not add them there. Not sure why but the triggers for this table are not in the database...so I did put them in this file so they get added when v-next runs.

PR #362, opened 2020-05-13
OCD-3293 fix: Remove length limit from pending CP Source Grouping

drbgfc:
OCD-3293


v-next.sql
```diff
@@ -1,3 +1,4 @@
+\i ocd-3293.sql
 \i ocd-3110.sql
```
andlar, line 2: These need to be removed. You should merge "staging" into your branch. It'll clean up the merge conflicts (temporarily anyway)

drbgfc: Addressed as requested in below commit, thanks

PR #366, opened 2020-06-18
OCD-2944 - Convert columns from acb name to acb id

tmy1313:
OCD-2944


ocd-2944.sql
```diff
@@ -0,0 +1,41 @@
...
+do $$
+begin
+	if exists (select * from information_schema.columns where table_schema = 'openchpl' and table_name = 'inheritance_errors_report'and column_name = 'acb') then
+		execute 'update openchpl.inheritance_errors_report ier set certification_body_id = (select certification_body_id from openchpl.certification_body cb where cb.name = ier.acb)';
+	end if;
+end
+$$;
+
+alter table openchpl.inheritance_errors_report
+drop column if exists acb;
```
andlar, line 14: I'm not sure of the "drop column" part. I think we've tended to do these kinds of things in two steps, where the first step does the data migration, and we have a second sql file that does the "removal" of columns. I'm not sure if that's actually necessary here, but wanted to open it for discussion

kekey1: I think it's okay the way it's done.. better to not need a second step later.

ocd-2944.sql
```diff
@@ -0,0 +1,41 @@
+-- Update inheritance_errors_report table
+alter table openchpl.inheritance_errors_report
+add column if not exists certification_body_id integer;
```
kekey1, line 3: Should this big "bigint" instead of "integer" to match what's in the table definition in the other file? Also need to add a NOT NULL constraint at some point later in this script. Same comment with the other table.

PR #367, opened 2020-06-26
update certified_product_summary view to handle status events correctly

tmy1313:
OCD-3307


v-next.sql
```diff
@@ -1,3 +1,5 @@
+\i ocd-3307.sql
```
andlar, line 1: You don't need to do this, or have the ocd-3307 file at all. If you're updating a view it is automatically replaced with line 4 of this file. View changes don't need anything in v-next or a specific file

PR #370, opened 2020-07-28
Ocd 3441 - Add RWT Eligibility Year to table and views

tmy1313:


ocd-3441.sql
```diff
@@ -0,0 +1,2 @@
+alter table openchpl.certified_product
+add column if not exists rwt_eligibility_year integer null;
```
andlar, line 2: If you're adding this column in here, doesn't it also need to be in the dev/openchpl_data_model.sql file, in the create script for the certified_product table?

dev/openchpl_views.sql
```diff
@@ -578,7 +579,8 @@ SELECT cp.certified_product_id,
        COALESCE(surv_closed.count_closed_surveillance_activities, 0::bigint) as closed_surveillance_count,
        COALESCE(nc_open.count_open_nonconformities, 0::bigint) AS open_nonconformity_count,
        COALESCE(nc_closed.count_closed_nonconformities, 0::bigint) AS closed_nonconformity_count,
-       surv_dates.surv_dates
+       surv_dates.surv_dates,
+       cp.rwt_eligibility_year
```
kekey1, line 583: I'm not sure if this column was added to the corresponding entity in the API. There are also two other views that select from certified_product (certified_product_summary and certified_product_search_result). I wonder if they need this new column too? If not, that's fine. The certified_product_search_result view is used for a user performing a search from the API - w/o the UI and collections pages. I don't know how often that gets tested (or used... ) but if we want to return the field in that API search it would need to be added.

PR #372, opened 2020-08-20
feat: Added  new CQM version and  added new versions to existing CQMs

palakpatel311:


andlar (dismissed):
The code is fine, but the commit message isn't right. Commit messages should be in the imperative / present tense, and to trigger them going into our RELEASE_NOTES need to have "!" at the end of the type of commit. This commit message should have been `data!: Add new CQM version and add new versions to existing CQMs`

The structure to follow is to fill in the phrase (skipping the "prefix" part) "If applied, this commit will <fill in the blank>" which would be "If applied, this commit will `Add new CQM version and add new versions to existing CQMs`"

It's not necessary to change the commit message now, and somewhat annoying, but worth thinking about in the future

tmy1313 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #375, opened 2020-10-01
OCD-3508 - Edit SVAP information for specific criteria in listing details

tmy1313:
OCD-3508


ocd-3508.sql
```diff
@@ -0,0 +1,57 @@
+DROP TABLE IF EXISTS openchpl.certification_result_svap;
+DROP TABLE IF EXISTS openchpl.svap_criteria_map;
+DROP TABLE IF EXISTS openchpl.svap;
+
+CREATE TABLE openchpl.svap (
+    id bigserial NOT NULL,
+    regulatory_text_citation varchar(30) NOT NULL,
```
andlar, line 7: Might want to check on expected values for this thing. Not sure what the max length is going to be

andlar: Might want to check on max length for theses values. Not sure what it will be

andlar (changes requested):
We need updates to the dev/data-model.sql and dev/something-triggers.sql files too

kekey1 (dismissed):

ocd-3508.sql
tmy1313: The largest in length in the spreadsheet is 13 chars.  I used 30 chars since that it what we used for the `criterion.number`.

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

andlar (changes requested):
Has merge conflicts

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

andlar (dismissed):
Approving this PR, but the ticket can't move to the next stage without a pending PR that removes this dummy data. We won't incorporate that pending PR yet, but we need to have it teed up to be used when it's required

changes/README.md
```diff
@@ -1,2 +0,0 @@
-## Need to add some directions on how the updated process works
```
andlar: This file shouldn't be deleted

andlar (changes requested):
I think we need an update to the soft-delete file. There's a section in there where certification_result entities are soft-deleted on trigger, so the mapping here should probably be there too

Also, and maybe this is a separate conversation, but we need a PR to remove the dummy data at some point. I don't know when that will be implemented, but we'll need it eventually, and it's easier to put it in place and let it hang out now, while it's fresh in our minds, than to try to remember it later

andlar (approved):

kekey1 (approved):

PR #382, opened 2020-12-15
OCD 3574: Add new CQM

palakpatel311:


v-next.sql
```diff
@@ -1,3 +1,8 @@
+\i ocd-3519.sql
+\i ocd-3554.sql
+\i ocd-3410.sql
+\i ocd-3574.sql
```
kekey1, line 1: This should only show one addition to this file and it should be your ocd-3574.sql file. Only tickets in STG should have ocd-* files here and any ticket that's been deployed to prod should not be here.

kekey1 (changes requested):

kekey1 (approved):

tmy1313 (approved):

PR #385, opened 2021-01-19
Ocd-3563 - Add table to support Listing Validation Report

tmy1313:


changes/ocd-3563.sql
```diff
@@ -0,0 +1,18 @@
+DROP TABLE IF EXISTS openchpl.listing_validation_report;
+
+CREATE TABLE openchpl.listing_validation_report (
+  id                        bigserial NOT NULL,
+  chpl_product_number       text NOT NULL,
```
kekey1, line 6: What do you think about storing this in the table even though it can change with listing updates? Would it be better to just use the certified product id as a foreign key? I'm asking because maybe it doesn't matter but it depends how often the job is going to run to calculate the errors. If it runs daily then maybe it's okay to store this string but if it's less frequently then I feel like we could run into inconsistencies with the chpl product number changing over time.

changes/ocd-3563.sql
```diff
@@ -0,0 +1,18 @@
...
+  product_name              text NOT NULL,
```
kekey1: Same question for product_name and certification_status_name, I guess. I get that it's faster to store it than the join the information from the details view but I'm just wondering about it being wrong as listings are updated.

dev/openchpl_data-model.sql
```diff
@@ -3419,6 +3419,20 @@ CREATE TABLE openchpl.pending_certified_product_measure_criteria (
       ON UPDATE CASCADE ON DELETE RESTRICT
 );
 
+CREATE TABLE openchpl.listing_validation_report (
+  id                        bigserial NOT NULL,
+  chpl_product_number       text NOT NULL,
+  product_name              text NOT NULL,
+  certification_status_name text NOT NULL,
+  error_message             text NOT NULL,
+  report_date               timestamp with time zone NOT NULL DEFAULT NOW(),
```
andlar: Do we have a standard for how we do timestamps? I don't know if we do, or should, or I just don't remember, but didn't we just do a mass conversion of some timestamps to "timestamp without time zone". Would it make sense to use that here? Or do I have it backwards?

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (approved):

andlar (approved):

PR #390, opened 2021-02-03
OCD-3517 - Add table certification_criterion_attribute and initial values

tmy1313:


kekey1 (changes requested):
Can you untangle the certified_product_upload commits from this?

PR #391, opened 2021-02-04
feat: add new table certification_criterion_attribute

tmy1313:
OCD-3517


kekey1 (dismissed):

andlar (dismissed):

changes/ocd-3517.sql
```diff
@@ -0,0 +1,42 @@
...
+  CONSTRAINT certification_criterion_attribute_pk PRIMARY KEY (id),
+  CONSTRAINT certification_criterion_id_fk FOREIGN KEY (criterion_id)
+        REFERENCES openchpl.certification_criterion (certification_criterion_id)
+        MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+);
+
+CREATE TRIGGER certification_criterion_attribute_audit AFTER INSERT OR UPDATE OR DELETE on openchpl.certification_criterion_attribute FOR EACH ROW EXECUTE PROCEDURE audit.if_modified_func();
+CREATE TRIGGER certification_criterion_attribute_timestamp BEFORE UPDATE on openchpl.certification_criterion_attribute FOR EACH ROW EXECUTE PROCEDURE openchpl.update_last_modified_date_column();
+
+INSERT INTO openchpl.certification_criterion_attribute (criterion_id, svap, last_modified_user) VALUES (21, true, -1);
```
andlar, line 20: If this runs twice, will it insert multiple rows allowing a given criteria to have SVAP? Isn't that the same issue we ran into with the dummy-data insert scripts? I know it isn't likely to run twice in PROD, but we have a policy of not doing this kind of insert in general, right?

andlar (changes requested):

changes/ocd-3517.sql
tmy1313: The first line in this script drops the table, so we won't have dups. 

We were getting dups with the svap stuff, because the creation of the table was in the ocd-????.sql file and the creation of the data was in the dummy-data.sql file.  So, when the ocd-???? file went away, we were only running the inserts with the drop table that was in the ocd-????.sql table.

andlar (approved):

kekey1 (approved):

PR #392, opened 2021-02-09
OCD-3609: Add missing mappings between removed g1/g2 measures and cures criteria

kekey1:


changes/ocd-3609.sql
```diff
@@ -0,0 +1,138 @@
...
+				WHERE criteria_id = origCriterionId
+				AND value = uploadValue)
+			);
+		
+		IF measureId IS NULL THEN
+			RAISE NOTICE 'No measure id found for criterion id % and upload value %', origCriterionId, uploadValue;
+		ELSE 
+			-- insert an entry for the cures criterion into the "deprecated" macra measure table
+			INSERT INTO openchpl.macra_criteria_map
+				(criteria_id, value, name, description, removed, last_modified_user)
```
andlar, line 23: I don't see any of the calls to this function passing "false" in for this value. Are _all_ of these measures/mappings/whatever actually removed?

andlar (dismissed):
I have no idea if the data is correct, but I don't see anything to be concerned with as to the functionality. Assuming my question on `removed` doesn't mean anything needs to change then I'm good with this one

changes/ocd-3609.sql
kekey1: Oh, yes they are all removed. The problem was with removed macra measures. I could have not had that parameter but put it in for whatever reason... I don't think it's hurting anything for this temporary function.

andlar: Works for me. Thanks!

tmy1313 (approved):

andlar (approved):

PR #397, opened 2021-02-18
OCD-3571 - Remove Spring 'acl_' tables

tmy1313:


kekey1 (dismissed):

changes/ocd-3571.sql
```diff
@@ -0,0 +1,5 @@
+DROP TABLE IF EXISTS openchpl.acl_entry;
```
andlar, line 1: Are there triggers that need to be removed?

dev/openchpl_data-model.sql
```diff
@@ -3499,6 +3384,7 @@ CREATE TABLE openchpl.certification_criterion_attribute (
   CONSTRAINT certification_criterion_id_fk FOREIGN KEY (criterion_id)
         REFERENCES openchpl.certification_criterion (certification_criterion_id)
         MATCH SIMPLE ON UPDATE NO ACTION ON DELETE RESTRICT
+);
```
andlar, line 3387: Where is this coming from?

tmy1313: For some reason I saw it was missing.  Not sure where/when it happened, but figured I'd just fix it.

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (commented):
Is that create-user function automatically dropped/reloaded in some of our load scripts or does it need to be dropped and reloaded in the ocd sql file that drops the tables?

kekey1 (approved):

andlar:
> Is that create-user function automatically dropped/reloaded in some of our load scripts or does it need to be dropped and reloaded in the ocd sql file that drops the tables?

The first line of the create-user.sql function drops it, and at least the load script that loads from maintenance executes that sql script before running the function that inserts the users. I think it's close enough to correct that we can test it out and if it's not working somewhere, we'll need to reconfigure something in our build pipeline, as opposed to the code itself

PR #409, opened 2021-05-03
OCD-3594: Remove background jobs tables

kekey1:


andlar (changes requested):
Don't we need to remove these from the `dev/data-model.sql` file as well?

andlar (approved):

tmy1313 (approved):

PR #420, opened 2021-07-13
OCD-3572 - Add new NON_CONFORMITY_CLOSE_DATE field to CHPL

tmy1313:


dev/openchpl_data-model.sql
```diff
@@ -2011,6 +2011,7 @@ CREATE TABLE openchpl.surveillance_nonconformity (
 	certification_criterion_id bigint,
 	nonconformity_type varchar(1024),
 	nonconformity_status_id bigint not null,
+	non_conformity_close_date date not null,
```
kekey1: It doesn't look like this is set to "NOT NULL" in the ocd-3572.sql file but i'm not sure if it needs to be removed from here or added there.

kekey1 (changes requested):

kekey1 (dismissed):

dev/openchpl_data-model.sql
```diff
@@ -2011,6 +2011,7 @@ CREATE TABLE openchpl.surveillance_nonconformity (
 	certification_criterion_id bigint,
 	nonconformity_type varchar(1024),
 	nonconformity_status_id bigint not null,
```
andlar, line 2013: Is this column still needed? This is a derived field now, right? So should we be dropping it here?

tmy1313: The way the ticket was written and the checkboxes that existed when I started, I think I am supposed to create a second ticket that will permanently remove the non-conformity status "stuff" from the DB.

kekey1 (approved):

andlar (approved):

PR #444, opened 2021-11-11
OCD-3803 - Update Optional Standards mapping and values

tmy1313:


changes/ocd-3803.sql
```diff
@@ -0,0 +1,7 @@
+update openchpl.optional_standard
+set description = 'CCDS: SNOMED CT® U.S. Edition, September 2019 Release'
```
andlar, line 2: Does the symbol after the "CT" show up correctly in all of the various places? XML, HTML, email (if these are in any emails), etc.?

kekey1 (approved):

changes/ocd-3803.sql
tmy1313: It works in the HTML and here is in the XML.  Looks good in my viewer.  Not sure if it appears in any emails.
image

andlar (approved):

PR #452, opened 2022-02-18
OCD-3751: Deprecate bulk reject of pending listings

kekey1:
[#OCD-3751]


andlar (dismissed):

tmy1313 (dismissed):

dev/openchpl_data-model.sql
```diff
@@ -854,7 +854,7 @@ CREATE TABLE openchpl.test_data_criteria_map (
 		MATCH FULL ON DELETE RESTRICT ON UPDATE CASCADE
 );
 
-CREATE TABLE openchpl.conformance_method_criteria_map (
+CREATE TABLE openchpl.certification_result_conformance_method (
```
andlar, line 857: I don't understand what's happening here. With the changes in this file it looks like the `conformance_method_criteria_map` table is going away, and there are two places where the `certification_resutl_conformance_method` table is created, though they have different columns. Given the changes in the ocd-3751 file, I would think that only one of these tables should be changed, and they should be changed in a way where whatever `version character varying(50) not null` became `version character varying(50)`, which I don't see

andlar (changes requested):

dev/openchpl_data-model.sql
kekey1: Yeah, I messed up something. Should be fixed.

andlar (approved):

tmy1313 (approved):

PR #471, opened 2022-06-13
OCD-3949 - Add script to truncate the shared_store.shared_store table

tmy1313:
OCD-3949


dev/reset-shared-store.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
andlar, line 3: There's a line in most of our scripts that's needed to make these scripts work in a cygwin environment, due to differences in EOL stuff:

```suggestion
(set -o igncr) 2>/dev/null && set -o igncr; # this comment is required to trick cygwin into dealing with windows vs. linux EOL characters

```

dev/reset-shared-store.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ $# -lt 2 ]; then
+    psql -Upostgres -c "TRUNCATE shared_store.shared_store;" openchpl
```
andlar: I don't like using the postgres user on CHPL tables. We have both the openchpl_dev and the openchpl users as defined users in our tables, but don't have access to the postgres one on our servers. I prefer having our local environments match the deployment environments as much as possible, but am not willing to die on this hill

dev/reset-shared-store.sh
```diff
@@ -0,0 +1,23 @@
...
+
+    psql -h $host -U $user -c "TRUNCATE shared_store.shared_store;" $db
+else
+    echo 'This script can be run with 0, 2, or 3 arguments. Please check your input and try again.'
+fi
+
+
+
+
+
```
andlar: Any reason for these lines?

kekey1 (dismissed):

andlar (dismissed):

andlar (approved):

kekey1 (approved):

PR #482, opened 2022-08-16
Ocd 3940 - Revise logic for "Associated ONC-ACBs" on change requests to be determined by the type of change request

tmy1313:


kekey1 (changes requested):
In the soft_delete file, there is a developer soft delete function that seems to also be marking related change requests as deleted. I think it may also need a change for the new table in this PR.

kekey1 (approved):

andlar (approved):

PR #493, opened 2022-10-11
OCD-4029 - Update API architecture for Surveillance domain

tmy1313:


changes/ocd-4029.sql
```diff
@@ -0,0 +1,200 @@
+create table if not exists openchpl.additional_nonconformity_type(
+    id bigint not null default nextval('openchpl.certification_criterion_certification_criterion_id_seq'),
```
kekey1, line 2: One thing that makes me slightly nervous about using this particular sequence is that we have a service that we use everywhere that gets criteria by ID. And we have lots of properties that reference criteria by ID. I know none of the existing criteria IDs will change and I understand why it makes sense to use this same sequence, but I assume there will be a day where we want to add new criteria, and we will run into issues with their IDs changing throughout dev/qa until the criteria are in production. I don't think it's a dealbreaker, but I suspect it will annoy us greatly when that time comes.

changes/ocd-4029.sql
```diff
@@ -0,0 +1,200 @@
...
+insert into openchpl.additional_requirement_detail_type(surveillance_requirement_type_id, name, removed, last_modified_user) 
+select 5, 'Semiannual Attestations Submission', false, -1
+where not exists (
+    select * 
+    from openchpl.additional_requirement_detail_type
+    where name = 'Semiannual Attestations Submission'
+    and removed = false
+);
+
+--If this has already been added, we need to delete it.
```
kekey1: Do you still need this delete statement?

kekey1 (dismissed):

kekey1 (dismissed):

andlar (dismissed):
Nothing to add to the other comments

kekey1 (dismissed):

andlar (dismissed):

kekey1 (approved):

andlar (approved):

PR #503, opened 2022-11-30
OCD-4092 - Remove unused view aggregated_nonconformity_statistics

tmy1313:
feat: remove unused view aggregated_nonconformity_statistics

### DIFF
--- a/dev/openchpl_views.sql
+++ b/dev/openchpl_views.sql
@@ -1378,20 +1378,3 @@ FROM openchpl.certified_product cp
          AND cse.certification_status_id IN (1, 6, 7)) as listing_status
          ON cp.certified_product_id = listing_status.certified_product_id
 AND cp.certification_edition_id = 3;
-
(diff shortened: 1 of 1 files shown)


kekey1 (approved):

dev/openchpl_views.sql
```diff
@@ -1378,20 +1378,3 @@ FROM openchpl.certified_product cp
          AND cse.certification_status_id IN (1, 6, 7)) as listing_status
          ON cp.certified_product_id = listing_status.certified_product_id
 AND cp.certification_edition_id = 3;
-
-CREATE OR REPLACE VIEW openchpl.aggregated_nonconformity_statistics
```
andlar, line 1382: Removing the view needs a "drop if exists" command somewhere, right?

kekey1: That is run at the beginning of this file, the line already exists. Should be removed at some point in the future.

andlar (approved):

PR #516, opened 2023-01-20
OCD-4107: ICS Errors Report table updates

kekey1:


changes/ocd-4107.sql
```diff
@@ -0,0 +1,26 @@
...
+
+ALTER TABLE openchpl.inheritance_errors_report 
+ADD CONSTRAINT certified_product_fk 
+	FOREIGN KEY (certified_product_id) 
+	REFERENCES openchpl.certified_product (certified_product_id) 
+	MATCH FULL ON UPDATE CASCADE ON DELETE RESTRICT;
+
+-- Allow URL field to be null and we won't fill it in any longer
+ALTER TABLE openchpl.inheritance_errors_report
+ALTER COLUMN url DROP NOT NULL;
```
andlar: What _was_ this column used for? Is there some functionality that's going away?

andlar (dismissed):

changes/ocd-4107.sql
kekey1: We were storing the URL to the listing in the database and putting it into the report. I changed it so that the database stores the certified_product_id and the report emailing job builds the URL with that ID + a property we already have. There are several other columns I considered deleting and just deriving in the report emailing job but did not. Can you try to remember to ask about this either at the FR or maybe we could just have a developer chat to see if you guys think it's important to remove  any others?

andlar (approved):

tmy1313 (approved):

PR #563, opened 2023-12-20
OCD-4373 - Populate all historical listings with Standards

tmy1313:


andlar (dismissed):

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+				raise notice 'Inserted Certification Result Standard %  -- %  -- %', v_cert_result.chpl_product_number, v_certification_criterion.number, v_standard.regulatory_text_citation;
+			else
+				raise notice 'Certification Result Standard Exists %  -- %  -- %', v_cert_result.chpl_product_number, v_certification_criterion.number, v_standard.regulatory_text_citation;
+			end if;
+		end loop;
+	end loop;
+end $$;
+
+
+call openchpl.backfill_standards('170.202(a)(2)', 'Direct Project: ONC Applicability Statement for Secure Health Transport, Version 1.2 August 2015', 3, '2016-01-14', '2016-01-14', null, null, 'Aplies to both b1 and b1 cures', array[16, 59, 60, 165]);
```
kekey1:
```suggestion
call openchpl.backfill_standards('170.202(a)(2)', 'Direct Project: ONC Applicability Statement for Secure Health Transport, Version 1.2 August 2015', 3, '2016-01-14', '2016-01-14', null, null, 'Applies to both b1 and b1 cures', array[16, 59, 60, 165]);
```

kekey1: Also, in the spreadsheet that was attached to the email you recently sent out, the "Additional Information" about applying to b1 and b1Cures doesn't appear to be attached to the h1 and h2 criteria. At least, they are different rows in the spreadsheet. Does the Additional Information matter for our purposes?

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+
+call openchpl.backfill_standards('170.202(b)', 'ONC XDR and XDM for Direct Messaging Specification', 3, '2016-01-14', '2016-01-14', null, null, null, array[60]);
+
+call openchpl.backfill_standards('170.202(d)', 'ONC Implementation Guide for Direct Edge Protocols, Version 1.1, June 25, 2014', 3, '2016-01-14', '2016-01-14', null, null, null, array[16, 60, 165]);
```
kekey1, line 113: In the spreadsheet where this Standard is applied to b1/Cures, it has two rules listed (2015 and Cures). Does that matter? It's just lost here, because we can't apply a Standard to 2 rules in our data model so I want to confirm that's fine. I actually wonder if a Standard should be applied to a rule at all because it's connected to criteria and whatever rules the associated criteria are connected with are the rules the standard is actually connected with. Now I kind of feel like it's incorrect to link a Standard and a Rule to begin with. If we have to pick one rule are we just picking the earliest one in time? So, we would favor 2015 over Cures? Maybe that's what we're doing here but the spreadsheet says "2015/Cures" which makes it look like the Standard is associated with 2 Rules. Thoughts?

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+
+call openchpl.backfill_standards('170.202(e)(1)', 'Delivery Notification - Implementation Guide for Delivery Notification in Direct v1.0', 3, '2016-01-14', '2016-01-14', null, null, null, array[40, 59, 60, 178]);
```
kekey1: Unless I'm misunderstanding the spreadsheet, which is very possible, 170.202(e)(1) is only associated with criteria IDs 59 and 60. NOT 40 and 178.

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+
```
kekey1, line 118: Related to the above comment, I think we are missing a line for 170.204(a)(1). Row 10 in the spreadsheet.

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+call openchpl.backfill_standards('170.210(g)', 'The date and time recorded utilize a system clock that has been synchronized using any Network Time Protocol (NTP) standard.', 3, '2016-01-14', '2016-01-14', null, null, null, array[30, 31, 38, 40]);
+
+call openchpl.backfill_standards('170.210(h)', 'Audit log content. ASTM E2147-18 Standard Specification for Audit and Disclosure Logs for Use in Health Information Systems.', 4, '2020-06-30', '2020-06-30', null, null, null, array[173, 174, 175]);
+
+call openchpl.backfill_standards('170.213(a)', 'United States Core Data for Interoperability (USCDI), July 2020 Errata, Version 1', 4, '2020-06-30', '2022-12-31', '2025-12-31', 'USCDI', 'Must be replaced by United States Core Data for Interoperability (USCDI), October 2022 Errata, Version 3', array[47, 165, 166, 178, 47, 180, 181, 182, 210]);
```
kekey1:
```suggestion
call openchpl.backfill_standards('170.213(a)', 'United States Core Data for Interoperability (USCDI), July 2020 Errata, Version 1', 4, '2020-06-30', '2022-12-31', '2025-12-31', 'USCDI', 'Must be replaced by United States Core Data for Interoperability (USCDI), October 2022 Errata, Version 3', array[47, 165, 166, 178, 180, 181, 182, 210]);
```

changes/ocd-4373.sql
```diff
@@ -0,0 +1,323 @@
...
+Identifies as Male. 446151000124109
+Identifies as Female. 446141000124107
+Female-to-Male (FTM)/Transgender Male/Trans Man. 407377005
+Male-to-Female (MTF)/Transgender Female/Trans Woman. 407376001
+Genderqueer, neither exclusively male nor female. 446131000124102
+Additional gender category or other, please specify. nullFlavor OTH
+Choose not to disclose. nullFlavor ASKU', 3, '2016-01-14', '2016-01-14', '2025-12-31', 'SOGI', 'Sexual Orientation and Gender Identity must be coded in accordance with, at a minimum, the version of SNOMED CT® codes specified in §170.207(a)(1)', array[5]);
+
+
+call openchpl.backfill_standards('170.210(e)(1)', 'The audit log must record the information specified in sections 7.1.1 and and 7.1.2 and 7.1.6 through 7.1.9 of the standard specified in § 170.210(h) and changes to user privileges when health IT is in use.
```
kekey1: Is criteria ID 174 left out on purpose? It's in the spreadsheet Row 139 but it's yellow and I don't know what yellow means

kekey1 (changes requested):

changes/ocd-4373.sql
tmy1313: This row has a significantly different Standard description.  It is noted in the Notes column that it needs to be confirmed by ONC.  I left it out,  until we get confirmation.  The way the script is written, we can add the criteria to the array and the data will be inserted, without any other data being (a)effected.

changes/ocd-4373.sql
tmy1313: The rule does not have anything to do what criteria it is associated with.  It is related to when the Standard was created.  Also, we don't use the data for anything, as of now.

changes/ocd-4373.sql
tmy1313: I removed the text from the Additional Info column - I think it was in the wrong column.

changes/ocd-4373.sql
kekey1: Ok, as long as there is a plan to follow up on it. Might be good to let QA note that so they don't find the same missing criteria.

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (dismissed):

andlar (dismissed):

kekey1 (dismissed):

kekey1 (dismissed):

andlar (dismissed):

kekey1 (dismissed):

andlar (approved):

kekey1 (approved):

PR #566, opened 2024-01-17
OCD-4437 - Create new report for listings that need to be updated for HTI-1

tmy1313:


andlar (approved):

changes/ocd-4437.sql
```diff
@@ -0,0 +1,30 @@
+create table if not exists openchpl.updated_listing_status_report (
+    id bigserial not null,
+    certified_product_id bigint not null,
+    report_day date not null,
+    criteria_require_update_count int,
+    days_updated_early int,
+    chpl_product_number text not null,
```
kekey1, line 7: Do we need to save this if we are saving the certified product id? Asking because this (and the columns below - product, version, developer, certification status) can change over time. If there is a reason to save this data "as it was" at the time the report data generator ran then I'm good with this. But if we wanted the emailed report to reflect this data "as it is now" then maybe we don't save all this stuff? Thoughts?

kekey1 (approved):

changes/ocd-4437.sql
kekey1: Discussed with Todd and learned that the report is intended to be a snapshot of the data as it was. Resolving the comment.

PR #583, opened 2024-04-01
OCD-4413 - Update Activity and Questionable Activity to incorporate Cognito users

tmy1313:


dev/openchpl_views.sql
```diff
@@ -1286,8 +1288,9 @@ AS
 	) all_questionable_activity
 	JOIN openchpl.questionable_activity_trigger trigger ON all_questionable_activity.questionable_activity_trigger_id = trigger.id
 	LEFT JOIN openchpl.listing_search ON all_questionable_activity.certified_product_id = listing_search.certified_product_id
-	JOIN openchpl.user u on all_questionable_activity.activity_user_id = u.user_id
-	JOIN openchpl.contact c on u.contact_id = c.contact_id;
+	--JOIN openchpl.user u on all_questionable_activity.activity_user_id = u.user_id
+	--JOIN openchpl.contact c on u.contact_id = c.contact_id
+	LEFT JOIN openchpl.activity act on all_questionable_activity.activity_id = act.activity_id;
```
kekey1, line 1289: My only concern with this is that not all questionable activity was mapped to an activity_id. There are thousands of questionable activities where we could not or did not make this historical mapping. I did the work of adding the activity_id mapping and trying to fill it in with OCD-4160 and was looking at the code for that but I still don't remember why some/many activity ids were not filled in. Many were filled in, as well. Anyway, maybe this needs more thought because if there is not activity id linkage it looks like we won't have the user id for some historical questionable activities the way this is changing.

dev/openchpl_views.sql
```diff
@@ -1219,14 +1219,16 @@ AS
 	all_questionable_activity.activity_date, all_questionable_activity.reason, 
 	all_questionable_activity.certification_status_change_reason,
 	all_questionable_activity.activity_id, 
-	all_questionable_activity.activity_user_id, 
+	--all_questionable_activity.activity_user_id,
```
kekey1: Remove the commented out lines if we don't need them

kekey1 (changes requested):

dev/openchpl_views.sql
kekey1: I also don't think we even tried to fill in the activity ID for non-listing questionable activities (developers, products, etc). There are a much smaller number of those obviously. OCD-4160 looks like it went out around a year ago (mid-April, 2023) so all questionable activities of all kinds are guaranteed to have an activity_id mapping from that date onward. Maybe we just need to ask if this is acceptable to not have the user for activities older than mid-April 2023?

kekey1 (approved):

andlar (approved):

PR #695, opened 2026-06-17
data!: Add metadata for Surveillance Activities ONC report

kekey1:
[#OCD-5335]


changes/ocd-5335.sql
```diff
@@ -0,0 +1,135 @@
+-- Another ticket removes use of the report_key column, so we should just allow nulls
```
andlar, line 1: Where is the ticket that removes the column? If that ticket goes first, will this command fail? Or just be ignored?

andlar (dismissed):

changes/ocd-5335.sql
kekey1: I ended up not removing the column with these tickets because I didn't see how I could know which would go first. It would technically be a breaking change, so I'm planning on removing the column when I remove the deprecated field from the API in some future ticket.

andlar (approved):

